Issue #26, opened 2019-04-27
How to convert a MatrixXdual to MatrixXd? 


AzureViolin:
I'm expecting something like:

```
MatrixXdual mat_dual;
MatrixXd mat_d = val(mat_dual);
```

However apparently val() can only be applied to dual, rather than any container of dual. Is there an existing way to do that?

Also, it took me quite a while to figure out how to do jacobian on multi-vector-variable function and conversion between dual and double, the readme should be updated. I can help if you want.

allanleal:
Hi. For the conversion between `dual` and `double` matrices, I think you can rely on the `cast` method of Eigen. Could you please try:

~~~c++
MatrixXdual mat_dual;
MatrixXd mat_double = mat_dual.cast<double>();
~~~

and let me know?

> Also, it took me quite a while to figure out how to do jacobian on multi-vector-variable function and conversion between dual and double, the readme should be updated. I can help if you want.

That would be great! Just open a PR when you can. Thanks a lot in advance.

AzureViolin:
Finally got time to get back to this, sorry for being late. 
I've tried cast back then and it doesn't work or have too much restrictions, can't really recall the exact issue with cast. I've created dedicated functions to solve my problem. Probably too rudimentary to be added into your implementation.  
```
//convert MatrixXdual to MatrixXd
MatrixXd mat_dual2d(const MatrixXdual& mat_dual){
  MatrixXd mat_d;
  for (size_t j = 0, nRows = mat_dual.rows(), nCols = mat_dual.cols(); j < nCols; ++j)
    for (size_t i = 0; i < nRows; ++i)
      mat_d(i,j) = val(mat_dual(i,j));
  return mat_d;
}

VectorXd vec_dual2d(const VectorXdual& vec_dual){
  size_t nRows = vec_dual.rows();
  VectorXd vec_d(nRows);
  //cout<<"vec_dual rows "<<nRows<<endl;
  for (size_t i = 0; i < nRows; ++i)
    vec_d(i) = val(vec_dual(i));
  return vec_d;
}
```

AzureViolin:
Regarding the multi variable jacobian, I don't know how to add a new item to your tutorial. 
[https://autodiff.github.io/tutorials/#jacobian-of-a-vector-function](url)
Specifically, I'd like to add a section after:
> Jacobian of a vector function

And call it

> Jacobian of a multi-vector function

[...]

PR #38, opened 2019-06-26
Fix for issue "How to convert a MatrixXdual to MatrixXd?" #26

ludkinm:
See comments in https://github.com/autodiff/autodiff/issues/26

### DIFF
--- a/autodiff/forward/forward.hpp
+++ b/autodiff/forward/forward.hpp
@@ -458,8 +458,8 @@ struct Dual
 
     Dual() : Dual(0.0) {}
 
-    operator T() const { return this->val; }
-    
+    explicit operator T() const { return this->val; }
+
     template<typename U, enableif<isNumber<U>>...>
     Dual(U&& other) : val(other), grad(0) {}
 
--- a/test/forward.test.cpp
+++ b/test/forward.test.cpp
@@ -724,4 +724,24 @@ TEST_CASE("autodiff::dual tests", "[dual]")
             for(auto j = 0; j < 2; ++j)
                 REQUIRE( J(i, j) == approx(-F[i] + ((i == j + 1) ? 1.0 : 0.0)) );
     }
+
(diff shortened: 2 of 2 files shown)


autodiff/forward/forward.hpp
```diff
@@ -458,6 +458,8 @@ struct Dual
 
     Dual() : Dual(0.0) {}
 
+    operator T() const { return this->val; }
```
allanleal: Thanks for the PR. I think an `explicit` conversion is required:

~~~c++
explicit operator T() const { return this->val; }
~~~

to avoid "accidental" conversions to `T` by the compiler, and thus spoiling the propagation of derivatives.

Let's see if the tests pass.

ludkinm: Thanks, that seems sensible.

allanleal: I think it would be more conservative if we go `explicit` here. I'll make this change.

allanleal: Could you please check if using `explicit` conversion works fine with Eigen on your side?

ludkinm: It seems to be working ok on some test cases. 
I can't run the full tests in ./autodiff/test. I've only been using the headers from the library thus far.

allanleal: OK. What I actually meant is that if the issue #26 is resolved with `explicit` conversion operator, instead of implicit. I cannot test it right now. Actually, it would make sense to have a test there. Can you give me a simple example, please, that shows the casting combined with Eigen usage?

ludkinm: I've added two tests to test/forward.test.cpp showing vector and matrix casts.
They pass locally (seems to have passed Travis too).

allanleal:
Great! Thanks @ludkinm , for also adding the tests!

PR #39, opened 2019-07-02
Fix dividing by zero in pow functions

supersega:
When evaluate derivetive of pow in zero point it is NaN


allanleal:
Thanks for the fix!

Issue #43, opened 2019-07-03
Problems Installing and linking with Eigen


dmckinnell:
None of the CMakeList files contain a command to find the Eigen package so the final installation command fails unless the files are altered. The example files that use the Eigen directories have an issue in the include command #include <eigen3/Eigen/Core> should just be #include <Eigen/Core>.

allanleal:
I see. I'll see if I can fix this now, adding `find_package(Eigen REQUIRED)` and propagate Eigen dependency to targets linking against `autodiff`.

allanleal:
Note that `eigen3/Eigen/Core` is the path one would use if `conda` is being used to install autodiff or if Eigen was installed using its `cmake` + `make install`.

pariterre:
I was about to open the same issue. 
I think referencing directly ```eigen3/Eigen/Core``` as the path is not the proper way to do since FindEigen is supposed to give you the path to ```eigen3``` already. Therefore the include should read ```Eigen/Core```.

allanleal:
@pariterre Thanks for the explanation and apologies for the delay (I'm on vacation). What you say makes sense then to change `eigen3/Eigen/Core` occurrences to `Eigen/Core`. I'll see if I can do it in the next few days - would you or @dmckinnell be able to kindly resolve this? Not necessarily using the PR #44 I opened that is still unfinished.

allanleal:
@pariterre Just saw you proposed a PR! Thanks.

allanleal:
OK, just merged PR #45 of @pariterre and created a new release v0.5.6 which should soon be available in conda-forge. Please reopen with necessary.

PR #44, opened 2019-07-03
Improve Eigen dependency resolution

allanleal:
This PR aims to improve dependency resolution for Eigen by propagating Eigen dependency in autodiff to other targets (executables/libraries). This is a result of the discussion in #43 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,5 +23,5 @@ add_subdirectory(autodiff)
 add_subdirectory(examples)
 add_subdirectory(test)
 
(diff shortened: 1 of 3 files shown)


ibell:
Personally, I never rely on the system package of Eigen.  I much prefer to add it as a submodule in git, and link to it directly.  This is partly because I like to be 100% sure that I have a version of Eigen that works, and who knows what is installed.

PaulXiCao:
The changes seem to be inline with the recommended way from [Eigen's documentation](https://eigen.tuxfamily.org/dox/TopicCMakeGuide.html). Is there more work needed? Testing?

PR #49, opened 2019-08-07
[Proposal] Make gradient vector and Jacobian matrix computation more general

supersega:
This is proposal (not finaly completed yet) of more general computition of gradient vector and Jacobian matrix computation. 
Through this implementation we can compute gradient for function with next signature:
```cpp
dual f(const VectorXdual& x, dual a, dual b, dual c)
{
    ...
}
```
Now we are able to call gradient wrt (a, b, c, x)
```cpp
 VectorXd g2 = gradient(f, wrt(a, b, c, x), at(x, a, b, c), u);
```

P.S.
Btw, I know why tests fail, and how to resolve this

### DIFF
--- a/autodiff/forward/eigen.hpp
+++ b/autodiff/forward/eigen.hpp
@@ -112,14 +112,21 @@ auto gradient(const Function& f, Wrt&& wrt, Args&& args, Result& u) -> Eigen::Ve
     detail::for_each(wrt, [&n](auto&& arg){ n += arg.size(); });
 
     Eigen::VectorXd g(n);
-
-    for(std::size_t j = 0; j < n; ++j)
+    
+    detail::for_each(wrt, [f, &args, &u, &g, index_aligment = std::size_t(0)](auto&& arg) mutable
     {
-        std::get<0>(wrt)[j].grad = 1.0;
-        u = std::apply(f, args);
-        std::get<0>(wrt)[j].grad = 0.0;
-        g[j] = u.grad;
-    }
+        std::size_t size = arg.size();
+
+        for (std::size_t j = 0; j < size; ++j)
+        {
+            arg[j].grad = 1.0;
+            u = std::apply(f, args);
+            arg[j].grad = 0.0;
+            g[j + index_aligment] = u.grad;
+        }
+
+        index_aligment += size;
+    });
 
     return g;
 }
--- a/autodiff/forward/eigen.hpp
+++ b/autodiff/forward/eigen.hpp
@@ -122,28 +122,71 @@ struct is_eigen_dual_vector<T, std::enable_if_t<is_eigen_dual_container_v<T>>> :
 template <typename T>
 inline constexpr bool is_eigen_dual_vector_v = is_eigen_dual_vector<T>::value;
 }
+
+/// Evaluate derivative of multi variable function at point args by 'wrt' variable arg 
+template<typename Function, typename Args, typename Arg, typename Result>
+void evaluate(const Function& function, Args&& args, Arg&& arg, Result& u)
+{
+    arg.grad = 1.0;
+    u = std::apply(function, args);
+    arg.grad = 0.0;
+}
+
+/// Helper class for computing gradient vector for different type of arguments
+/// Example of function signature : VectorXdual f(VectorXdual x, VectorXdual y, dual p) 
+template <typename Function, typename Args, typename Gradient, typename Result>
+struct gradient_filler
+{
+    gradient_filler(Function f, Args args, Gradient& g, Result& u) :
+        f(f), args(args), g(g), u(u), index_aligment(0) { }
+    
+    /// Fill gradient vector for duals vector
+    template <typename Arg, std::enable_if_t<detail::is_eigen_dual_vector_v<Arg>>...>
+    void operator()(Arg&& arg) 
+    {
+        for (std::size_t j = 0; j < arg.size(); ++j)
+        {
+            evaluate(f, args, arg[j], u);
+            g[j + index_aligment] = u.grad;
+        }        
+        index_aligment += arg.size();
+    }
+    
+    /// Fill gradient vector for single dual variable
+    template <typename Arg, std::enable_if_t<isDual<Arg>>...>
+    void operator()(Arg&& arg)
+    {
+        evaluate(f, args, arg, u);
+        g[index_aligment] = u.grad;
+        index_aligment += 1;
+    }
+private:
+    Function f;
+    Args args;
+    Gradient& g;
+    Result& u;
+    std::size_t index_aligment;
+};
+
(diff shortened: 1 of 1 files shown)


allanleal:
Hi @supersega 

What about this solution instead:

~~~c++
dual f(const VectorXdual& x, dual a, dual b, dual c)
{
    ...
}

auto g = [](const VectorXdual& xabc)
{
    const auto n = xabc.size();
    const auto x = xabc.head(n - 3);
    const auto abc = xabc.tail(3);
    const auto a = abc[0];
    const auto b = abc[1];
    const auto c = abc[2];
    return f(x, a, b, c);
}
~~~

and then:

~~~c++
VectorXd g2 = gradient(g, wrt(xabc), at(xabc), u);
~~~

supersega:
@allanleal Yes, I know this solution. But we need to create this wrappers in client code. For each kind of signature.

allanleal:
The problem is that `wrt(x, y, z)` is currently intended to compute higher-order derivatives, e.g., ∂³f/∂x∂y∂z.

supersega:
Yes, I’ve already seen usage of wrt. We can introduce new function to avoid confusion. If this is the only problem.

allanleal:
Better to change it then for higher order diff. 

We then have `wrt(x, a, b, c)` with the meaning you're suggesting, since it makes a good parallel with `at(x, a, b, c)`. And for higher order diff, we could use `wrt({x, y, z})` or just change the function name, say, `xwrt`, with `x` of extended. 

My intention is to have a different and simpler way (more efficient also) to compute higher order diff in the future. Either with order determined at compite time (for higher performance) or at runtime (for higher flexibility).

supersega:
What you think about use in case with gradient `wrt_pack` name?

allanleal:
I have thought about your suggestion (`wrt_pack`) and looked again at the examples. 

For `gradient` and `jacobian` evaluations, using forward diff, we rely on Eigen for input and output types. Your use case mixes vector and scalar variables as inputs (`x` a vector, `a`, `b`, `c` scalars). The following:

~~~c++
VectorXdual gx = gradient(f, wrt(x), at(x, a, b, c)); 
~~~

would compute only the gradient with respect to vector `x`. The usage below is (currently) not supported:

~~~c++
VectorXdual ga = gradient(f, wrt(a), at(x, a, b, c)); // ga = [ df/da ]
VectorXdual gb = gradient(f, wrt(b), at(x, a, b, c)); // gb = [ df/db ]
VectorXdual gc = gradient(f, wrt(c), at(x, a, b, c)); // gc = [ df/dc ]
~~~

and I'm not sure it should be supported (for scalar variables), since one could use the method `derivative` instead. I know, this is not what you are proposing in this PR.

What you want is to be able to compute `g = [ df/dx, df/da, df/db, df/dc ]` in a convenient way. I was about to question why you don't put all these variables in an extended vector, but just recalled that I have a use case in which inputs are temperature (*T*), pressure (*P*), and a vector of moles of chemical elements/species (**b** or **n**), and the scalars temperature and pressure are kept separately (*T*, *P*, **b**). Thus, I might eventually also need a vector `[ df/dT, df/dP, df/db ]`! :)

I would say however that we preserve things as much as possible and introduce a function/command `pack` (based on your suggestion) that can be combined with `wrt` in the following way:

~~~c++
dual f(const VectorXdual& x, dual a, dual b, dual c)
{
    ...
}

VectorXd g = gradient(f, wrt(pack(x, a, b, c)), at(x, a, b, c), u); // g = [ df/dx, df/da, df/db, df/dc ]
~~~

One could even ignore `b` and `c` derivatives, for example:

~~~c++
VectorXd g = gradient(f, wrt(pack(x, a)), at(x, a, b, c), u); // g = [ df/dx, df/da ]
~~~

Not sure how difficult it would be to implement this. Maybe `wrtpack` would be simpler (`wrt_pack` would break the method naming conventions adopted, which avoids snake cake for public methods, but for variables it's ok).

supersega:
Hi @allanleal, Wow, that realy good comments, I think that
```cpp
VectorXd g = gradient(f, wrt(pack(x, a)), at(x, a, b, c), u);
```
or
```cpp
VectorXd g = gradient(f, wrt(variables(x, a)), at(x, a, b, c), u);
```
looks more 'naturally' than 
```cpp
VectorXd g = gradient(f, wrtpack(x, a), at(x, a, b, c), u);
```
Not sure what should return `pack` or `variables`, probably it will by just "dummy" function which will improve code readability.

Btw, in my case I need to calculate Jacobian for boundary value problem continuation, and I have function with parameters
```cpp
auto ode_system(const VectorXdual& state_variables, dual frequence)
{
...
}
```
And previously I created wrapper via lambda, but then I think it should work from the box.
So, what you think about dummy function? =)

allanleal:
Have a look at lines 98-114 of file `autodiff/forward/eigen.hpp`:

~~~c++
/// Return the gradient vector of scalar function *f* with respect to some or all variables *x*.
template<typename Function, typename Wrt, typename Args, typename Result>
auto gradient(const Function& f, Wrt&& wrt, Args&& args, Result& u) -> Eigen::VectorXd
{
    const std::size_t n = std::get<0>(wrt).size();

    Eigen::VectorXd g(n);

    for(std::size_t j = 0; j < n; ++j)
    {
        std::get<0>(wrt)[j].grad = 1.0;
        u = std::apply(f, args);
        std::get<0>(wrt)[j].grad = 0.0;
        g[j] = u.grad;
    }

    return g;
}
~~~

Here, the argument `wrt` of template type `Wrt` is currently expected to be a tuple with:

1. one or more items
2. these items are expected to be vectors (see the need for `operator[j]`)

However, note that it only process the first item of the tuple `wrt` (check the calls `std::get<0>(wrt)`). 

We need to change this code so that:

1. size variable `n` is the sum of sizes of each item in argument `wrt`
2. the for loop is replaced with an *outer loop* (one for each item in `wrt`) and an *inner loop* (with `std::get<0>` replaced by `std::get<i>` where `i` is the iteration number of the inner loop).

So, the `pack` function should convert scalar values of `dual` type (`Dual<T,G>` more precisely) to a wrapper type that has methods `size()` (which returns 1) and operator [i] (which return the value it self). I think these are all the requirements of the template code above, if I'm not overlooking anything.

This is a lot of steps; let me know if I was not clear enough. ;)

supersega:
@allanleal I've already implemented foreach for tuple, so we can use it ;)

allanleal:
> @allanleal I've already implemented foreach for tuple, so we can use it ;)
> P.S. I enjoy your idea about wrapping Dual

Excellent, the compile-time foreach for tuples is what we need! ;)

supersega:
Yesterday I was experimenting on wrapping duals, and I’ve found the solution. So, today or tomorrow I’ll create new PR.

PR #50, opened 2019-08-17
Make grad and jac more general

supersega:
Hi @allanleal, as we discussed before #49 this PR makes computation of gradient vector and Jacobi matrix more general. I think it will be great if we add examples with new functionality.


allanleal:
Hi @supersega ; sorry for the delay. I was busy with conference and I'm now back into the office. I'll see if I can have a look at this today or tomorrow. Thanks a lot for your work.

test/forward.test.cpp
```diff
@@ -750,4 +752,118 @@ TEST_CASE("autodiff::dual tests", "[dual]")
...
+        x << 1.0, 2.0, 3.0;
+        
+        dual y = 2;
+   
+        VectorXdual z(4);
+        z << 1.0, 2.0, 3.0, 4.0;
+
+        VectorXd g = gradient(f, wrtpack(x.tail(2), y, z), at(x, y, z));
+        
+        REQUIRE( g.size() == 7 );
```
allanleal: Could you also please test the gradient values?

autodiff/forward/eigen.hpp
```diff
@@ -165,4 +273,7 @@ auto jacobian(const Function& f, Wrt&& wrt, Args&& args) -> Eigen::MatrixXd
 
 } // namespace autodiff::forward
 
+namespace autodiff {
+using forward::wrtpack;
```
allanleal: We'll probably have just `wrt` in the future, when the usage of `wrt` to compute higher-order derivatives is modified. We may want to have such note in the tutorials.

test/forward.test.cpp
```diff
@@ -750,4 +752,118 @@ TEST_CASE("autodiff::dual tests", "[dual]")
...
+
+        VectorXdual z(3);
+        z << 1.0, 2.0, 3.0;
+
+        VectorXdual F;
+
+        const MatrixXd J = jacobian(f, wrtpack(x.tail(2), y), at(x, y, z), F);
+
+        REQUIRE(J.rows() == 6);
+        REQUIRE(J.cols() == 3);
```
allanleal: Here as well; the function is relatively simple. It should be easy for you to test the calculated derivatives against their expected values.

autodiff/forward/eigen.hpp
```diff
@@ -94,21 +94,108 @@ EIGEN_MAKE_TYPEDEFS_ALL_SIZES(autodiff::dual, dual)
...
+constexpr auto tailView(Tuple&& t) {
+    constexpr auto size = std::tuple_size<std::remove_reference_t<Tuple>>::value;
+    static_assert(N <= size, "N must be smaller or equal than size of tuple");
+    return viewImpl(t, makeIndexSequence<size - N, size>{});
+}
+
+/// Wrap T to compatible array interface
+template<typename T>
+struct EigenVectorAdaptor {
+    /// implicit construct from value
```
allanleal: Cool; exactly what we needed.

test/forward.test.cpp
supersega: Not sure that it is necessary, since they calculated in next test case.

autodiff/forward/eigen.hpp
supersega: Ok, I'll do examples.

test/forward.test.cpp
supersega: Not sure that it is necessary, since they calculated in next test case.

supersega:
Hi @allanleal. Sorry for the delay. Please see my answers about your review.

test/forward.test.cpp
allanleal: Sorry, I missed this email from GitHub. OK, makes sense then not to have it.

autodiff/forward/eigen.hpp
```diff
@@ -54,6 +54,23 @@ template<> struct NumTraits<autodiff::dual> : NumTraits<double> // permits to ge
     };
 };
 
+template<> struct NumTraits<autodiff::HigherOrderDual<2>> : NumTraits<double> // permits to get the epsilon, dummy_precision, lowest, highest functions
```
allanleal: Good to see you back @supersega ! :)
Is it possible to make this generic with respect to the order (instead of 2, a `std::size_t N`)? Not sure if necessary though, if we just want to use eigen up to 2nd order to compute Hessian matrices.

autodiff/forward/eigen.hpp
```diff
@@ -271,6 +306,54 @@ auto jacobian(const Function& f, Wrt&& wrt, Args&& args) -> Eigen::MatrixXd
     return jacobian(f, std::forward<Wrt>(wrt), std::forward<Args>(args), F);
 }
 
+/// Return the hessian matrix of scalar function *f* with respect to some or all variables *x*.
+template<typename Function, typename Wrt, typename Args, typename Result, typename Gradient>
+auto hessian(const Function& f, Wrt&& wrt, Args&& args, Result& u, Gradient& g) -> Eigen::MatrixXd
```
allanleal: Nice!

autodiff/forward/eigen.hpp
supersega: Hi @allanleal same to you. Probably yes, but I'm not sure. We can improve code quality in nex PR's, since this one is too big now (IMHO).

autodiff/forward/eigen.hpp
supersega: Thanks :+1:

supersega:
@allanleal I have a few ideas and want to discuss.

allanleal:
Let's use gitter to discuss, maybe hangouts for chatting. 

https://gitter.im/autodiff/community

autodiff/forward/eigen.hpp
supersega: Hi @allanleal , sorry for the delay(again), I've tied to avoid duplication of code using templates, but it looks like we can use the only macro here. So, let merge this PR, and then make new things based on it =)

allanleal:
Hi @supersega ; You asked me to merge a PR last week. I'm wondering if it is this one? I'm dedicating this week to this project, and I'll be able to review/merge this and other PRs. Sorry for not acting earlier on this.

supersega:
Hi @allanleal yes, it is. Btw do you have some new ideas? What if we add some numerical methods based on autodiff forward?

allanleal:
> Hi @allanleal yes, it is. Btw do you have some new ideas? 

So, the new forward algorithm I was implementing for the past two days (for higher-order derivatives and non-recursive, compile-time optimized expressions) is only suitable for higher order directional derivatives. For example, to generate Taylor series for either scalar or vector functions along a given direction. So, I think we should stick with the way `dual` is implemented now (using a recursive algorithm and recursive data structure, e.g., `dual2nd` is `Dual<Dual<double,double>, Dual<double,double>>`), which allows higher-order cross derivative calculations very easily. The drawback is that a `dual10th` type would require 1024 doubles!

For the new non-recursive implementation, I'll define a new forward diff type called `real`. The use of `real` will then be optimal for first or higher order directional derivatives. I know, it's a bit vague and I'm not giving examples here. 

@ibell Would directional derivatives be useful to you? Are you. like me, computing derivatives of substance properties? :) I colleague told me you develop coolprop. 

  > What if we add some numerical methods based on autodiff forward?

Yes, we could start with a non-linear solver that computes the Jacobian automatically if the residual function is implemented using `dual`. Other solvers could be implemented later on, such as non-linear minimization algorithms, ODEs, etc. But I think these could be a new project. Let's chat on gitter.about this.

supersega:
What about this PR?;)

ibell:
@allanleal Yes, I was/am the primary developer of CoolProp, but now I work at NIST, and develop REFPROP.  Many of the same problems :) The directional derivatives come in for me with calculating critical points, which require quite accurate directional derivatives.  And building them analytically is a nightmare, so AD is a lifesaver. For that problem, being able to calculate the gradient and Hessian matrix is basically the big challenge for me. I would vote to keep autodiff focused on being really good (and fast) with differentiation, and leave the higher-level things you can do with them to another package (in my opinion).  You might want to consider to write a short paper for JOSS about autodiff: https://joss.theoj.org/

At work I'm developing https://github.com/usnistgov/CEGO/ , which is primarily a stochastic optimization library, but I'm hitting the wall with stochastic optimization, so I'm looking now into hybrid stochastic/deterministic optimization, and for that, I need gradients and Hessians.

allanleal:
Hi Ian, I added you to the gitter channel, to have a more convenient discussion about plans and what we are doing and how autodiff can help us. Thanks for the journal recommendation; it does look like a good medium to publish about autodiff. The CEGO project looks quite nice! I too need to perform numerical optimization (minimization of Gibbs energy for chemical equilibrium calculations). I use a deterministic approach, based on interior-point algorithms. To compute the Hessian, I do not actually need second order derivatives, because *H* is the Jacobian of the vector of chemical potentials, and these can be computed directly. By directly I mean without differentiating the Gibbs energy function.  Maybe this could be something similar in your use case.

supersega:
@allanleal could you review please, this branch is very old;)

autodiff/forward/eigen.hpp
```diff
@@ -94,21 +124,116 @@ EIGEN_MAKE_TYPEDEFS_ALL_SIZES(autodiff::dual, dual)
...
+    Eigen::Index size() const {
+        return 1;
+    }
+private:
+    T val;
+};
+
+/// Meta function to endow floating point array compatible
+template<typename T>
+using makeTypeCompatible = std::conditional<isDual<std::decay_t<T>>, EigenVectorAdaptor<T>, T>;
```
allanleal: You should not need `std::decay` here because `isDual` already does this behind the scenes (removes cv, reference, volatile using `plain<T>` template alias).

allanleal:
```suggestion
using makeTypeCompatible = std::conditional<isDual<T>, EigenVectorAdaptor<T>, T>;
```

supersega: Yes, that’s true. I’ll fix it at the evening.

allanleal:
I have already done it! :)

CI is passing!

allanleal:
@supersega I'll merge this today or tomorrow, and then merge/adapt your contribution with the new version of autodiff that I'm finalizing. We'll have `wrt` acting equivalently to `wrtpack` when used with `gradient` and `jacobian` functions.

allanleal:
Thanks @supersega .

PR #52, opened 2019-09-05
Fix expression tree figure linking in README

volpatto:
Hi, Allan.

I was curious about autodiff and found a very minor problem. PR's reason is provided below:

This PR performs a very minor fix. In the autodiff's README from master branch, the expression tree figure is not properly loaded. This PR just update the link, pointing to the correct place.


allanleal:
Thanks a lot, Diego!

Issue #61, opened 2019-09-18
Location of VectorXdual


ibell:
In my opinion, the ``VectorXdual`` typedef doesn't belong in the Eigen namespace, I think it would be more sensible to put it in the ``autodiff`` namespace.  It's hard to see this from the examples because the ``using namespace ...`` paradigm is used which obfuscates where things are coming from.

allanleal:
I agree with you. Would you kindly do a pull request with these changes,
and also the ArrayXdual typedefs?

ibell:
Sure!  Happy to do so.  Autodiff is a game changer for me

allanleal:
Great Ian. Your help is much appreciated.

allanleal:
This issue was fixed with PR #63

Issue #60, opened 2019-09-18
Can't add to VectorXdual


ibell:
On MSVC 2019, this code
``` C++
// C++ includes
#include <iostream>

// autodiff include
#include <autodiff/forward.hpp>
#include "Eigen/Dense"
#include <autodiff/forward/eigen.hpp>
using namespace autodiff;

int main()
{
    Eigen::VectorXdual x(2);  // the input vector x
    x << -0.5, 0.5;
    Eigen::VectorXdual x2 = x + 0.5 * 0.001;
    std::cout << x2 << "\n";
}
```
yields the compile error
```
1>D:\Code\helmderiv\snippets\bugdualvec.cpp(14,31): error C2676:  binary '+': 'Eigen::VectorXdual' does not define this operator or a conversion to a type acceptable to the predefined operator
```

ibell:
Sorry, my fault (sort of).  Eigen::Vector don't allow for coefficient-wise addition, but Eigen::Array do.  Could we add Eigen::Array typedefs too for generality?  Of course also possible to do two conversions, one to array, and another back to matrix.  But that's unnecessarily verbose.

ludkinm:
@ibell 
In this case you can do
```c++
 Eigen::VectorXdual x2 = x.array() + 0.5 * 0.001;
```
Although, be careful. Using the ``.array`` member function won't work inside expressions that take derivatives.

allanleal:
> Using the .array member function won't work inside expressions that take derivatives.

Hi @ludkinm , what do you mean above? Could you please give an example showing what should be the expected result and the actual one when using method `.array()` of Eigen on vectors/matrices?

ludkinm:
I meant it would not compile.
Say f(x) = sum( x*(x+1) )
```c++
return x.cwiseProduct(x.array() + 1.0).sum();
```
However, I realise if you construct the result into VectorXdual then it works:
```c++
return x.cwiseProduct(VectorXdual(x.array() + 1.0)).sum();
```
In my previous comment I guess the constructor for ``VectorXdual`` from ``VectorX`` is being called (explicitly), whereas ``VectorXdual::cwiseProduct(VectorXdual x)`` expects a ``VectorXdual``.
Since, the ``VectorXdual::array()`` returns a some kind of Eigen array object this is not converted to VectorXdual since there is no conversion operator (?)

PR #63, opened 2019-09-19
Add ArrayXdual typedefs, move into autodiff namespace

ibell:
See #61 
See #60

Did I handle this correctly?  The tests all run without any modifications, but now the typedefs are in the autodiff namespace.  Should I have put them in autodiff::forward?  That was borking compilation of the tests.


allanleal:
Thanks @ibell .

Issue #57, opened 2019-09-07
N-th derivative


ibell:
Working through the examples in the tutorial (very nice, by the way), I can see how to hard-code the construction of the ``wrt`` argument for first- and second-order derivatives of a one-dimensional (and multivariate) function, and that works fine.  But when it comes to higher derivatives, I've had no luck.  I've tried every combination and permutation of ``std::apply``, perfect forwarding, and so on that I can think of, but I have to think that this might just be a current limitation of how argument forwarding is handled in ``wrt()``?  If not, I would be very happy to be wrong.  I feel like I'm very, very close to getting this working.  Of course I can hard-code the higher derivatives (as I will do for now), but I don't want to have to do that.

``` C++
// C++ includes
#include <iostream>
#include <chrono>
#include <iomanip>
using namespace std;

// autodiff include
#include <autodiff/forward.hpp>
using namespace autodiff;

/*
Calculate the n-th derivative of the (one-dimensional) function sin(x)
*/
template <std::size_t n> double deriv(double x)
{
    using vartype = HigherOrderDual<n>;
    vartype _x = x; // explicit copy construction (makes the shared_ptr)
    std::array<vartype, n> args; args.fill(_x); // fill copy constructor
    return derivative(
        [](vartype x)->vartype{ return sin(x); }, 
        std::apply(wrt, args), 
        at(_x)
    );
}

int main()
{
    std::cout << deriv<1>(0) << std::endl;
}
```
yields
```
[ 50%] Building CXX object CMakeFiles/Nderiv.dir/snippets/Nderiv.cpp.o
/home/ian/Code/helmderiv/snippets/Nderiv.cpp: In instantiation of ‘double deriv(double) [with long unsigned int n = 1]’:
/home/ian/Code/helmderiv/snippets/Nderiv.cpp:28:28:   required from here
/home/ian/Code/helmderiv/snippets/Nderiv.cpp:21:19: error: no matching function for call to ‘apply(<unresolved overloaded function type>, std::array<autodiff::forward::Dual<double, double>, 1>&)’
         std::apply(wrt, args),
         ~~~~~~~~~~^~~~~~~~~~~
In file included from /usr/include/c++/7/bits/unique_ptr.h:37:0,
                 from /usr/include/c++/7/bits/locale_conv.h:41,
                 from /usr/include/c++/7/locale:43,
                 from /usr/include/c++/7/iomanip:43,
[...]

PR #64, opened 2019-09-23
[autodiff/forward] Simplify higher order derivatives

supersega:
Add new wrt function overloading. As we discussed #57 users want to use simpler way of higher-order derivatives calculation. I remember that @allanleal thought about changing dual implementation, but I'm not sure that it is necessary and easy to implement. Also, we will get regressions using such an approach. 
```cpp
using dual4th = Dual<double, 4>;

dual4th x = 1.0;
dual4th y = derivative(sin, wrt(x), at(x));

cout << y[0] << endl; // value of the 0th order derivative of sin(x), which is its actual value
cout << y[1] << endl; // value of the 1st order derivative of sin(x)
cout << y[2] << endl; // value of the 2nd order derivative of sin(x)
cout << y[3] << endl; // value of the 3rd order derivative of sin(x)
cout << y[4] << endl; // value of the 4th order derivative of sin(x)
```

Here we calculate 4 derivatives and it's mean that we call sin 4 times, what if the user wants only one in some cases? 

Consolations: We can change the implementation of Dual to avoid recursion, but we should not add overhead to the user, and that's mean that we need wrt in that form which we have =) correct me if I'm wrong.

P.S. FYI @ibell

### DIFF
--- a/autodiff/forward/forward.hpp
+++ b/autodiff/forward/forward.hpp
@@ -591,6 +591,21 @@ auto seed(Arg& dual, Args&... duals) -> void
     dual.grad = num;
 }
 
+template<typename T>
+constexpr auto repeat(T&& t, std::index_sequence<0>) 
+{
+    // Just stop recursion
+    return std::forward_as_tuple(std::forward<T>(t));
+}
+
+template<typename T, std::size_t I, std::size_t... N>
+constexpr auto repeat(T&& t, std::index_sequence<I, N...>) 
+{
+    // concat tuple with rest N
+    return std::tuple_cat(std::forward_as_tuple(std::forward<T>(t)),
+        repeat<T>(std::forward<T>(t), std::make_index_sequence<sizeof...(N)>{}));
+}
+
 } // namespace internal
 
 template<typename Arg>
--- a/autodiff/forward/forward.hpp
+++ b/autodiff/forward/forward.hpp
@@ -640,6 +640,12 @@ auto wrt(Args&&... args)
     return std::forward_as_tuple(std::forward<Args>(args)...);
 }
 
+template<std::size_t N, typename Wrt>
+auto wrt(Wrt&& arg) 
+{
+    return internal::repeat<Wrt>(std::forward<Wrt>(arg), std::make_index_sequence<N>{});
+}
+
 template<typename... Args>
 auto at(Args&&... args)
 {
--- a/examples/forward/example-forward-higher-order-derivatives.cpp
+++ b/examples/forward/example-forward-higher-order-derivatives.cpp
@@ -24,10 +24,10 @@ int main()
     dual ux = derivative(f, wrt(x), at(x, y));  // evaluate the derivative du/dx
     dual uy = derivative(f, wrt(y), at(x, y));  // evaluate the derivative du/dy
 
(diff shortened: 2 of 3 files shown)


ibell:
What happens if you try to get an index out of range?

allanleal:
Hi @supersega . 

> Here we calculate 4 derivatives and it's mean that we call sin 4 times, what if the user wants only one in some cases?

We know that `d sin(x)/dx` is `cos(x)` and `d cos(x)/dx` is `-sin(x)`. These recurrences would be used in the implementation of higher order derivatives (10th order derivative of `sin(x)` would actually only require one evaluation of `std::sin` and one of `std::cos`, not 10). For `exp(x)`, for example, we would only require one evaluation of function `std::exp`; the same for `log(x)` and a few other mathematical functions. So, there would be no overhead here. In fact, the current recursive implementation causes memory overhead (`Nth` order requires `2**N` doubles; compare this with the `N+1` memory requirement of the new implementation I have in mind!) :)


Hi @ibell 

> What happens if you try to get an index out of range?

You would see a compile error because of the use of `std::array` to store the value and derivatives! If you actually mean using `wrt<4>(x)` in this PR proposal of @supersega , in which only up to 2 should be used, you'll most likely also get a compile error.

allanleal:
> Here we calculate 4 derivatives and it's mean that we call sin 4 times, what if the user wants only one in some cases?

Further clarifications. An Nth order derivative computation using automatic differentiation would also compute the other lower order derivatives. There is no overhead here either, because this is how you would compute them anyway manually. 

Check below how we would compute the derivatives of `exp(x)` up to *n*-th order. 

Things to note first: `x[0]` is the value of `x`, `x[i]` for `1 < i <= n`are seed values to control partial derivative computation.

image

Note above that to compute `exp(x)[i]` we evaluate first `exp(x)[i-1]`. Also note that `exp(x)[0]` is `std::exp(x[0])` and computed only once!

supersega:
Hi @allanleal thanks for explanation. Looks like I have some gaps in theory =) 
If I understand correctly, we should remake all expression nodes with compile time for loop?

allanleal:
> If I understand correctly, we should remake all expression nodes with compile time for loop?

What I mean is that those `n+1` statements in the figure above (for evaluating `exp` on a future/new `dual` type) would be expanded at compile-time, instead of a regular, runtime for loop.

supersega:
Yes, I see. Btw, how to deal with mul expression in this case?

allanleal:
image

Above, multiplication of `x` and `y`.

I have already done this analytical expansion for most mathematical operations; just need to implement them! :)

supersega:
Ah, now I see. If you had time, I would be glad to discuss it in evening(after office hours) in gitter.

allanleal:
Sure! There could indeed have a regression with this new implementation, but this would come with great benefits to future (and also current) users.

supersega:
@alanleal, sounds like we can merge this PR, since you leave class Dual ;)

allanleal:
@supersega Sounds good. Could you please just add this usage in one of the tutorial examples on computing higher-order diff? ;)

Otherwise even us can forget about this usage capability in the future!

supersega:
But I’ve already updated examples)

allanleal:
OK, of course. :)

allanleal:
Great, done! Thanks!

Issue #62, opened 2019-09-18
Unable to nest functions


ibell:
This one puzzles me, and I suspect it's a bug.  If I expand the calls to pow2(x), the code compiles and works, but if I don't, I get the error below:

``` C++
// C++ includes
#include <iostream>

// autodiff include
#include <autodiff/forward.hpp>
#include "Eigen/Dense"
#include <autodiff/forward/eigen.hpp>

template <typename T>
T pow2(const T x0) {
    return x0 * x0;
}

template <typename T>
T Rosenbrock(const T x0, const T x1) {
    return 100.0 * pow2(pow2(x0) - x1) + pow2(1 - x0);
}

int main()
{
    using namespace autodiff;
    Eigen::VectorXdual x(2);  // the input vector x
    x << -0.5, 0.5;
    auto Rosenbrockdual = [](const Eigen::VectorXdual& x) {
        return Rosenbrock(x[0], x[1]);
    };
    dual F;
    auto grad = gradient(Rosenbrockdual, wrt(x), at(x), F);
    std::cout << grad;
}
```
yields
```
1>------ Build started: Project: bugdualvec, Configuration: Debug x64 ------
1>bugdualvec.cpp
1>D:\Code\helmderiv\snippets\bugdualvec.cpp(12,1): error C2440:  'return': cannot convert from 'autodiff::forward::BinaryExpr<autodiff::forward::MulOp,L,R>' to 'autodiff::forward::BinaryExpr<autodiff::forward::AddOp,L,R>'
1>D:\Code\helmderiv\snippets\bugdualvec.cpp(12,1): error C2440:         with
1>D:\Code\helmderiv\snippets\bugdualvec.cpp(12,1): error C2440:         [
1>D:\Code\helmderiv\snippets\bugdualvec.cpp(12,1): error C2440:             L=const autodiff::forward::BinaryExpr<autodiff::forward::AddOp,autodiff::forward::Dual<double,double>,autodiff::forward::UnaryExpr<autodiff::forward::NegOp,const autodiff::forward::Dual<double,double> &>> &,
1>D:\Code\helmderiv\snippets\bugdualvec.cpp(12,1): error C2440:             R=const autodiff::forward::BinaryExpr<autodiff::forward::AddOp,autodiff::forward::Dual<double,double>,autodiff::forward::UnaryExpr<autodiff::forward::NegOp,const autodiff::forward::Dual<double,double> &>> &
[...]

PR #68, opened 2019-10-03
Fix nan and/or segfault when using nested template functions

allanleal:
This is a fix for issue #62 in which template expression objects were being stored in other expression objects by reference. 

However, the use case in issue #62 still needs to be corrected. 

Consider the following usage of `dual`:

~~~c++
#include <iostream>

#include <autodiff/forward.hpp>
using namespace autodiff;

template <typename T>
auto pow2(T arg)
{
    return arg * arg;
}

int main()
{
    dual x = 2.0;

    dual y = pow2(pow2(x));

    std::cout << y << std::endl;
}
~~~

This application outputs `nan` or produces a segmentation fault. The reason is that argument `arg` in `pow2` is passed by value, so a temporary `dual` is created when calling `pow2(x)` in `main`. 
This temporary `dual` is then captured by reference when creating the template expression node 
`BinaryExpr<MulOp, const dual&, const dual&>`. This expression node is then returned, the temporary destroyed, and a dangling reference issue takes place. 

To correct this usage, one only needs to make sure the template argument `arg` is passed by reference:

~~~c++
template <typename T>
auto pow2(const T& arg)
{
    return arg * arg;
}
~~~

However, if a temporary is passed to `pow2`, for example, `pow2(dual(2.0))`, then the most recommended approach is to used *universal references*:

~~~c++
template <typename T>
auto pow2(T&& arg)
{
    return std::forward<T>(arg) * std::forward<T>(arg);
}
~~~

This will ensure that not a reference to the temporary is stored in the expression object, but a copy itself: `BinaryExpr<MulOp, dual, dual>`.

@ibell You may want to have a look at the above explanation. Unfortunately, using template functions with `dual` needs to be done more carefully, to avoid dangling references. This is not an issue when `dual` numbers are explicitly stated as the return value of functions. When using template, this is not possible (because it destroys the generic purpose of the template function definition) and instead of returning a `dual` number, an expression tree/node object is returned instead, possibly containing references to temporaries in the scope of the function.

### DIFF
--- a/autodiff/forward/forward.hpp
+++ b/autodiff/forward/forward.hpp
@@ -607,14 +607,14 @@ auto seed(Arg& dual, Args&... duals) -> void
 }
 
 template<typename T>
-constexpr auto repeat(T&& t, std::index_sequence<0>) 
+constexpr auto repeat(T&& t, std::index_sequence<0>)
 {
     // Just stop recursion
     return std::forward_as_tuple(std::forward<T>(t));
 }
 
 template<typename T, std::size_t I, std::size_t... N>
-constexpr auto repeat(T&& t, std::index_sequence<I, N...>) 
+constexpr auto repeat(T&& t, std::index_sequence<I, N...>)
 {
     // concat tuple with rest N
     return std::tuple_cat(std::forward_as_tuple(std::forward<T>(t)),
@@ -656,7 +656,7 @@ auto wrt(Args&&... args)
 }
 
 template<std::size_t N, typename Wrt>
-auto wrt(Wrt&& arg) 
+auto wrt(Wrt&& arg)
 {
     return internal::repeat<Wrt>(std::forward<Wrt>(arg), std::make_index_sequence<N>{});
 }
--- a/autodiff/forward/forward.hpp
+++ b/autodiff/forward/forward.hpp
@@ -702,6 +702,14 @@ auto derivative(const Function& f, Wrt&& wrt, Args&& args)
 //
 //=====================================================================================================================
 
+/// Alias template used to prevent expression nodes to be stored as references.
+/// For example, the following should not exist `BinaryExpr<AddOp, const dual&, const UnaryExpr<NegOp, const dual&>&>>`.
+/// It should be instead `BinaryExpr<AddOp, const dual&, UnaryExpr<NegOp, const dual&>>`.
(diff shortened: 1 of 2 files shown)


ibell:
Thanks for fixing this, but my nervousness with this issue remains.

Is there a way to *detect* this problem and error out somehow?  The limitation that you have to never make a mistake with how you deal with function arguments and return values and templates is a pretty serious one in my opinion, as the error could be pretty well buried in templated user code.  We are all human after all, and that's why we have type guarantees and compilers, right :)  All it would take is to miss one single & and all the calculated values are wrong, with no warning whatsoever.  I get that this is a limitation of the implementation and of C++, but it makes me very nervous about using a library where one tiny mistake is fatal.  

Just a thought, could you perhaps delete the default copy constructor for dual and only accept const references? Or would that bork all the code?

allanleal:
This is pretty much a C++ deficiency, which does not have the concept of borrow checker like Rust, which can do a full static analysis of the lifetime of references and issue a compilation error instead of runtime segmentation fault.

One would face exactly the same issue with any other C++ library out there that relies on template expressions and reference storage to avoid temporary memory allocation (e.g. try to return an Eigen matrix-vector multiplication where the matrix and/or vector are allocated inside the function or are arguments passed by value).

This use case you have presented, of possibly writing an entire numerical library using generic number types (so that one could use double or dual or any other number type) is almost convincing me that **we should not generate template expressions when performing dual mathematical/arithmetical operations (something similar to Eigen)**. Instead, we should immediately compute the result as dual number. Since dual lives in the stack memory, and never issue dynamic memory allocation, I think there won't be any performance issue with the new implementation.

This change in implementation would solve this issue entirely. However, before taking this decision and wiping out all the template metaprogramming code there to generate the template expressions, I'm considering doing a benchmark. 

**My inclination is to NOT have template expressions for forward automatic differentiation algorithms in autodiff.**

supersega:
@ibell deleting of copy ctor doesn’t solve your problem. As @allanleal said we can solve this issue only if we avoid expression creation. And we should create benchmarks to avoid time regressions. 

Btw, I’ve created benchmarks with catch for reverse mode. And I‘ll fail PR soon. FYI @allanleal.

allanleal:
On a process of merging all these PRs that are being incorporated into another PR with major improvements and capability additions. Thanks for the contribution!

PR #69, opened 2019-10-04
Implement a new autodiff type called real for improved higher-order directional derivative calculations

allanleal:
This PR implements a new `autodiff` type called `real` (an alias for `Real<1, double>`). Other aliases are:

~~~c++
using real1st = Real<1, double>;
using real2nd = Real<2, double>;
using real3rd = Real<3, double>;
using real4th = Real<4, double>;
using real5th = Real<5, double>;
using real6th = Real<6, double>;
using real7th = Real<7, double>;
using real8th = Real<8, double>;
using real9th = Real<9, double>;
~~~

One can check current implementation by inspecting files `autodiff/forward/real.hpp` and the test file `tests/real.test.cpp`.

The `real` type is pretty much 100% implemented. It supports all arithmetic operators and the following mathematical functions:

~~~
abs acos acosh asin asinh atan atanh cos cosh exp log log10 pow sin sinh sqrt tan tanh 
~~~

What is missing is the implementation of `VectorXr` and the calculation of directional derivatives of functions `VectorXr f(const VectorXr& x)` along a given vector `VectorXr v`.


allanleal:
Closing and re-opening this PR in a moment.

supersega:
Trying to attach screenshots from phone.

environment.devenv.yml
```diff
@@ -6,6 +6,7 @@ dependencies:
   - eigen
   - catch2
   - gxx_linux-64=7.3.0  # [linux]
+  - clangxx_osx-64=9.0.0  # [osx]
```
volpatto: Allan, I think that maybe you can try `- compilers` instead. This is not a recommended approach, but there is a chance that this way works.

volpatto: However, I don't know how `autodiff`'s cmake is configured. If it's configured to use `clang`, so the above comment will not work.

volpatto: > However, I don't know how `autodiff`'s cmake is configured. If it's configured to use `clang`, so the above comment will not work.

Ok, laziness is not a good thing (unless in lazy evaluations). So I will break my laziness and have a look in cmake as soon as I have some time.

allanleal: I added this `clangxx_osx-64` to ensure that when compiling for macOS (e.g., in Travis), we use the same compiler that conda-forge is using. However, Travis is exporting `CXX` to its own compiler - I need to overwrite this.

How is this `- compilers` supposed to work?

volpatto: Ah, `- compilers` will get a full set of compilers compatible to your system _and_ conda. You can check it [here](https://anaconda.org/conda-forge/compilers), but unfortunately without details.

autodiff/common/binomialcoefficient.hpp
```diff
@@ -0,0 +1,113 @@
...
+    1, 50, 1225, 19600, 230300, 2118760, 15890700, 99884400, 536878650, 2505433700, 10272278170, 37353738800, 121399651100, 354860518600, 937845656300, 2250829575120, 4923689695575, 9847379391150, 18053528883775, 30405943383200, 47129212243960, 67327446062800, 88749815264600, 108043253365600, 121548660036300, 126410606437752, 121548660036300, 108043253365600, 88749815264600, 67327446062800, 47129212243960, 30405943383200, 18053528883775, 9847379391150, 4923689695575, 2250829575120, 937845656300, 354860518600, 121399651100, 37353738800, 10272278170, 2505433700, 536878650, 99884400, 15890700, 2118760, 230300, 19600, 1225, 50, 1
+};
```
volpatto: I think that you can easily generate this code with [Cog](https://nedbatchelder.com/code/cog/) with a few lines of Python code. What do you think? You could even use `scipy.special.binom`.

allanleal: Thanks for sharing this. I have a python script somewhere that generates this `binomialcoefficient.hpp` file for a given order. The pascal triangle is generated in a few lines of code. I did not know about Cog, but the script is simple (just need to find it and add it to the project).

environment.devenv.yml
allanleal: Cool, so I can remove those two lines and use `compilers` instead, right?

volpatto: Yeah, at least I think so!

autodiff/common/binomialcoefficient.hpp
volpatto: I can do it in another PR in the future, if you want. Cog is a nice tool, it fits great in such situations like this one.

autodiff/forward/real/real.hpp
```diff
@@ -0,0 +1,835 @@
...
+            return c * x[i - j] * expx[j];
+        });
+    });
+    return expx;
+}
+
+template<size_t N, typename T>
+constexpr auto log(const Real<N, T>& x)
+{
+    assert(x[0] != 0 && "autodiff: log(x) has undefined value and derivatives when x = 0");
```
supersega: It can be not only a strict zero. What do you think about checking small values?

autodiff/forward/real/real.hpp
```diff
@@ -0,0 +1,835 @@
...
+}
+
+template<size_t N, typename T, typename U, EnableIf<isNumber<U>>...>
+constexpr auto pow(const Real<N, T>& x, const U& c)
+{
+    Real<N, T> res;
+    res[0] = pow(x[0], static_cast<T>(c));
+    if constexpr (N > 0)
+    {
+        assert(x[0] != 0 && "autodiff: pow(x, y) has undefined derivatives when x = 0");
```
supersega: Actually, pow has derivative at zero (for c >= 1) could you give me a quick explanation? (Sorry, I have no much time to underlying math).

autodiff/forward/real/real.hpp
```diff
@@ -0,0 +1,835 @@
...
+}
+
+//=====================================================================================================================
+//
+// TRIGONOMETRIC FUNCTIONS
+//
+//=====================================================================================================================
+
+template<size_t N, typename T>
+auto sincos(const Real<N, T>& x) -> std::tuple<Real<N, T>, Real<N, T>>
```
supersega: Can we move it to a detail namespace? Since it is not good for the user.

autodiff/forward/utils/derivative.hpp
```diff
@@ -0,0 +1,230 @@
...
+template<typename... Args, typename... Vecs>
+auto seed(const At<Args...>& at, const Along<Vecs...>& along)
+{
+    static_assert(sizeof...(Args) == sizeof...(Vecs));
+
+    ForEach(at.args, along.args, [&](auto& arg, auto&& dir) constexpr {
+        if constexpr (isVector<decltype(arg)>) {
+            static_assert(isVector<decltype(dir)>);
+            assert(arg.size() == dir.size());
+            const size_t len = dir.size();
```
supersega: For modern compilers, we don't need to create this `len` variable. It should optimize loop
```cpp
for(size_t i = 0; i < dir.size(); ++i)
```

allanleal:
Thanks a lot, @supersega. I'll reply to your comments later today.

autodiff/common/classtraits.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+//==========================================================================================================================================================
+// The code below was taken from: https://stackoverflow.com/questions/87372/check-if-a-class-has-a-member-function-of-a-given-signature/16867422#16867422
+// It implements checkers to determine if a type has a member variable, function, etc.
+//==========================================================================================================================================================
+
+template <typename... Args> struct ambiguate : public Args... {};
+
+template<typename A, typename = void>
+struct got_type : std::false_type {};
```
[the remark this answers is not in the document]
allanleal: I got these codes/macros from this [stackoverflow thread](https://stackoverflow.com/questions/87372/check-if-a-class-has-a-member-function-of-a-given-signature/16867422#16867422).

I'm no longer using them though; I used them initially to determine if a type has a `size()` method, to distinguish between vector-like and number entries in the `wrt` list. 

I have introduced `VectorTraits` now (`vectortraits.hpp`) that better handles this usage.

autodiff/common/meta.hpp
```diff
@@ -0,0 +1,174 @@
...
+    // ALTERNATIVE IMPLEMENTATION POSSIBLY USEFUL TO KEEP IT HERE
+    // auto g = [&](auto&&... args) constexpr {
+    //     ( f(std::forward<decltype(args)>(args)), ...);
+    // };
+    // std::apply(g, std::forward<Tuple>(tuple));
+    //------------------------------------------------------------
+}
+
+template<typename Tuple1, typename Tuple2, typename Function>
+constexpr auto ForEach(Tuple1&& tuple1, Tuple2&& tuple2, Function&& f)
```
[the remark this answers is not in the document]
allanleal: Yes, in the method `seed(at, along)`, where both `at` and `along` are wrappers to tuples, and we need to walk along them simultaneously and use current direction value in `along` to seed the current autodiff number in `at`.

autodiff/common/vectortraits.hpp
```diff
@@ -0,0 +1,84 @@
...
+using VectorValueType = typename VectorTraits<PlainType<V>>::ValueType;
+
+/// A template alias used to get the type of a vector that is equivalent to another but with a different value type.
+template<typename V, typename NewValueType>
+using VectorReplaceValueType = typename VectorTraits<V>::template ReplaceValueType<NewValueType>;
+
+/// A compile-time constant that indicates with a type is a vector type.
+template<typename V>
+constexpr bool isVector = !std::is_same_v<VectorValueType<PlainType<V>>, VectorTraitsNotDefinedFor<PlainType<V>>>;
+
```
allanleal: Check this compile-time bool variable `isVector<V>` that is `true` if `V` satisfies vector traits.

autodiff/common/vectortraits.hpp
```diff
@@ -0,0 +1,84 @@
...
+    using ReplaceValueType = VectorReplaceValueTypeNotSupportedFor<V>;
+};
+
+/// A template alias used to get the type of the values in a vector type.
+template<typename V>
+using VectorValueType = typename VectorTraits<PlainType<V>>::ValueType;
+
+/// A template alias used to get the type of a vector that is equivalent to another but with a different value type.
+template<typename V, typename NewValueType>
+using VectorReplaceValueType = typename VectorTraits<V>::template ReplaceValueType<NewValueType>;
```
allanleal: Check this - say we have type `SomeVectorType<dual4th>`, and we want to produce an equivalent vector with the underlying floating point value of `dual4th`, that is, `SomeVectorType<double>`, we can use this template alias `VectorReplaceValueType`.

`VectorReplaceValueType<VectorXdual3rd> === VectorXd`

If `dual3rd` above was instead defined with `float` (`using dual3rd = HigherOrderDual<3, float>`), then we woud get `VectorXf` instead.

autodiff/forward/utils/derivative.hpp
allanleal: OK

autodiff/forward/real/real.hpp
allanleal: See the equation:

image

In the specific case *y* is a constant, and not a `Real<N, T>` number, then we can compute the derivative when `x = 0` and `c >= N`, where `N` is the order of the `Real` number. Otherwise, there will be (at some point) a negative exponent on `0`. 

I think we should consider `pow(x, y) === exp(y * log(x))`, independently if `x` or `y` are constants. Note, however, that we do not compute `pow(x, y)` with `exp(y * log(x))`.

autodiff/forward/real/real.hpp
allanleal: I think we now have 99% of all codes in a `detail` namespace and whatever is supposed to be public, we export to the parent namespace `autodiff`. Hopefully we'll be able to use modules in the next C++ version, C++20, which will make this less painful.

autodiff/forward/real/real.hpp
allanleal: I propose we stick with strict zero; otherwise we need to define what is a small value, which is not always intuitive or possible.

autodiff/forward/real/real.hpp
```diff
@@ -0,0 +1,835 @@
...
+using std::tanh;
+
+//=====================================================================================================================
+//
+// TYPE TRAITS
+//
+//=====================================================================================================================
+
+template<typename T>
+struct _isReal { constexpr static bool value = false; };
```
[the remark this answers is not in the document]
allanleal: This is an auxiliary structure used later to define the template alias `isReal`: 

~~~c++

template<typename T>
struct _isReal { constexpr static bool value = false; };

template<size_t N, typename T>
struct _isReal<Real<N, T>> { constexpr static bool value = true; };

template<typename T>
constexpr bool isReal = _isReal<PlainType<T>>::value;
~~~

These definitions are all already inside a `detail` namespace. We can either rename `_isReal` to `isRealAux` or put it inside another child namespace called `impl` (`autodiff::detail::impl`).

I think the `isSomethingAux` and `AuxSomething` naming conventions are simpler, but the `impl` namespace above is also neat.

autodiff/forward/real/real.hpp
```diff
@@ -0,0 +1,835 @@
...
+// autodiff includes
+#include <autodiff/common/binomialcoefficient.hpp>
+#include <autodiff/common/numbertraits.hpp>
+#include <autodiff/common/meta.hpp>
+
+namespace autodiff {
+namespace detail {
+
+/// The type used to represent a real number that supports up to *N*-th order derivative calculation.
+template<size_t N, typename T>
```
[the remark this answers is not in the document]
allanleal: I create this issue #86 to make sure we have this eventually. This needs to be done more carefully, because we need to permit also custom/user defined numbers to be used with `Real` and `Dual` (e.g., high-precision floating point number types).

ci/actions/install-autodiff-unix.sh
```diff
@@ -0,0 +1,10 @@
+export PATH=$HOME/miniconda/bin/:$PATH
+source activate autodiff
```
marcelotrevisani: you can use another approach here
you can do something like
```
source $HOME/miniconda/bin/activate
```
and after that you can use just ``conda activate autodiff``

This way is better because it will load the conda functions as well instead of just appending to the path

marcelotrevisani: The same is valid for your windows script

allanleal: @volpatto Have a look at this please.

volpatto: Yeah, actually I did some modifications in `autodiff`'s CI files. I would like to suggest to update this branch with `master`. Everything related with `CI`, when solving the conflicts, should be selected from `master`. Now we are using a built-in `miniconda` from GH Actions. Let's see what happens, I'm available to help in this regard.

danilomendesdias: FYI: I merged this branch with master (#100), but compilation problems are still there. @tadeu helped me with that and found a bug with the compiler (already reported in https://developercommunity.visualstudio.com/content/problem/902908/compilation-problem-when-using-stdapply-with-eigen.html).

allanleal: @tadeu Knows all about MSVC! :)

Is there a way to change to a newer version, or an earlier version? Best would be to test here with the same version of MSVC used in autodiff-feedstock.

tadeu: I think that the bug is also present in the latest version, so we are out of luck with updating (until they release a fixed version), and we'll have to try to find a workaround.

The problem is with `std::apply`, it is not able to automatically convert an `Eigen::Matrix` to an `Eigen::Array`, even though when calling the function without `std::apply`, it works fine, and the parameter conversion is automatic.

allanleal: Thanks @tadeu - this should help. Perhaps I can try to avoid the conversion somehow (that I was not aware of). This would then not require a bug fix for MSVC.

PR #70, opened 2019-10-05
[WIP] Improve reverse mode

supersega:
This PR should improve performance when using `autodiff::reverse` algorithm.
It's just a part of work. See the task list below.

- [x]  make allocation on the stack.
- [ ] make enable to allocate expression nodes using allocator.
- [ ] add benchmarking. 
- [ ] compare stack allocation and pool (on heap).

### DIFF
--- a/autodiff/reverse/memory.hpp
+++ b/autodiff/reverse/memory.hpp
@@ -1,0 +1,102 @@
+//                  _  _
+//  _   _|_ _  _|o_|__|_
+// (_||_||_(_)(_|| |  |
+//
+// automatic differentiation made easier in C++
+// https://github.com/autodiff/autodiff
+//
+// Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+//
+// Copyright (c) 2018-2019 Allan Leal
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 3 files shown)


supersega:
@allanleal could you have a look on this, I think it's better to review PR iterative.

supersega:
@allanleal Looks like compiler on CI doesn't support all C++ 17 features. GCC has this file in experimental from **6.1** and in regular from **9.1**. Clang in experimental namespace from **5.0.0** and in regular from **9.0.0**.

allanleal:
Looks like a good plan, @supersega !

I haven't used `memory_resource` before. I assume this is the "problematic" file you mention. I'll be reviewing this as you go. I like the iterative approach. It makes things easier when it comes to a mergeable state.

supersega:
Exactly, memory_resource is that file. Now I’m going to implement resource manager, which will be singleton. Not sure about multi threading(I have no experience in this).

ChrisDrozdowski:
A couple of things related to memory.hpp:
1. MSVC 15.9.12 complains about placement of `[[maybe_unused]] ` and fails to compile.
It wants the placement to be, for example:
```
void_t* do_allocate(std::size_t bytes, [[maybe_unused]] std::size_t alignment = alignof(std::max_align_t)) override
```
and
```
void do_deallocate([[maybe_unused]] void_t* pointer, [[maybe_unused]] size_t bytes, [[maybe_unused]] size_t alignment) noexcept override {
```

2. Should `do_allocate`, `do_deallocate`, and `do_is_equal` not be private? They work properly with MSVC 15.9.12 as public, but they are originally documented to be private in specs.

supersega:
Hi @ChrisDrozdowski, thanks for pointing this. 

1. I did this experiments on visual studio 2017, and actually I don’t remember update version. It was working for me. But I’ll have a look. 
2. Yes, they should be private. 

P.S. Please note, that it is experimental branch, and code here is not clean.

adam-ce:
> Exactly, memory_resource is that file. Now I’m going to implement resource manager, which will be singleton. Not sure about multi threading(I have no experience in this).

in my opinion this type of project should really not use singletons. at the very least they should be optional.

moreover, all user facing classes should be reentrant. introducing a singleton resource manager will produce a bottleneck if somebody wants to use threading on a larger machine (e.g. i have a 24 thread workstation, there are workstations with 128 threads and soon more).

I didn't review this patch, and I don't know autodiff in detail. putting things on the stack is certainly a good idea, using fewer indirections as well (but i don't know whether that's possible). I don't think memory allocation is something that this project should deal with, in form of a template parameter at most. I had to exchange the default allocator once, because it was fragmenting memory too much -- that would be way harder, if our libs would use their custom memory manager.

just my two cents, hope it helps :)

PR #76, opened 2019-11-01
Convert font object in the logo to object path

allanleal:
This is a fix to resolve logo font not installed in some systems.


allanleal:
@tadeu do you know why this `continuous-integration/travis-ci Expected — Waiting for status to be reported` don't finish? This is now under the same state for more than 8h.

PR #79, opened 2019-11-09
Actions badges for master and cron jobs

volpatto:
This PR changes three things: 

* An improvement in the status badges in the README. Now, the badges are not polluted by Pull Requests status. Instead, they will only show the status of the last build for `master` branch;

* Cron jobs: builds for `autodiff` will occur every Monday 05 o'clock. This way, we can know when something unexpected occurred, some dependency updates and etc. No surprises!

* Builds for each OS are now separated. Each build has its status badge. It's better this way, we can see where things are broken in a blink of eyes.


allanleal:
Thanks @volpatto . I'm merging this PR and others because they have been already merged in PR #69 that I plan to merge today or tomorrow.

Issue #81, opened 2019-11-11
Scalar operations on VectorXvar and VectorXd


ludkinm:
This results in Eigen compilation errors
```c++
#include "Eigen/Core"
#include "autodiff/reverse.hpp"
#include "autodiff/reverse/eigen.hpp"

int main(void){
  Eigen::VectorXvar Xv(10);
  Eigen::VectorXd Y(10);
  autodiff::var z = (Xv -Y).squaredNorm();
  return 0;
}
```

_Originally posted by @ludkinm in https://github.com/autodiff/autodiff/issues/73#issuecomment-551807003_

allanleal:
Thanks @ludkinm  - just merged your PR.

PR #82, opened 2019-11-11
added ScalarBinaryOpTraits for var and double

ludkinm:
This PR allows for mixing VectorXvar and VectorXd types in element-wise binary operations such as
```c++
VectorXvar x(3); x << 1.0, 2.0, 3.0;
VectorXd y(3); x << 2.0, 2.0, 2.0;
x+y;
x-y;
x*y;
x.cwiseQuotient(y);
```
This is a fix for issue #81


allanleal:
Thanks @ludkinm . I'll merge this now, and then in PR #69 (which might get merged today with major improvements, new type `real` (and also `real2nd`, ..., `realNth = Real<N, double>`), Taylor series, and bug fixes. There are some breaking changes though - but these should be easy to fix.

PR #84, opened 2019-11-12
Erf forwards

ludkinm:
This PR adds Erf to forwards mode with a test.

Note that Erf (and other special functions) in Eigen are UnaryOps which apply to array objects, e.g:
```c++ 
VectorXdual x; 
erf(x.array());
```
Some special math functions are in the Eigen header (cf [Eigen bugzilla](https://eigen.tuxfamily.org/bz/show_bug.cgi?id=1775)):
``
 #include <unsupported/Eigen/SpecialFunctions>
``

### DIFF
--- a/autodiff/forward/forward.hpp
+++ b/autodiff/forward/forward.hpp
@@ -60,6 +60,7 @@ using std::tan;
 using std::cosh;
 using std::sinh;
 using std::tanh;
+using std::erf;
 
 //=====================================================================================================================
 //
@@ -83,9 +84,9 @@ struct InvOp    {};  // INVERSE OPERATOR
 struct SinOp    {};  // SINE OPERATOR
 struct CosOp    {};  // COSINE OPERATOR
 struct TanOp    {};  // TANGENT OPERATOR
-struct SinhOp    {};  // HYPERBOLIC SINE OPERATOR
-struct CoshOp    {};  // HYPERBOLIC COSINE OPERATOR
-struct TanhOp    {};  // HYPERBOLIC TANGENT OPERATOR
+struct SinhOp   {};  // HYPERBOLIC SINE OPERATOR
+struct CoshOp   {};  // HYPERBOLIC COSINE OPERATOR
+struct TanhOp   {};  // HYPERBOLIC TANGENT OPERATOR
 struct ArcSinOp {};  // ARC SINE OPERATOR
 struct ArcCosOp {};  // ARC COSINE OPERATOR
 struct ArcTanOp {};  // ARC TANGENT OPERATOR
@@ -95,6 +96,7 @@ struct Log10Op  {};  // BASE-10 LOGARITHM OPERATOR
 struct SqrtOp   {};  // SQUARE ROOT OPERATOR
 struct PowOp    {};  // POWER OPERATOR
 struct AbsOp    {};  // ABSOLUTE OPERATOR
+struct ErfOp    {};  // ERROR FUNCTION OPERATOR
 
 //-----------------------------------------------------------------------------
 // OTHER OPERATORS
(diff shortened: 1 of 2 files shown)


ludkinm:
There is an install fail. I think its due to the Eigen "unsupported" header.
I don't use conda so can't check this at my end easily.

test/forward.test.cpp
```diff
@@ -738,6 +750,24 @@ TEST_CASE("Eigen::VectorXdual tests", "[dual]")
         REQUIRE( g[2] == approx(x[2]) );
     }
 
+    SECTION("testing gradient derivatives with Erf")
+    {
+        auto f = [](const VectorXdual& x) -> dual
+        {
+            return erf(x.array()).sum();
```
allanleal: So, there is no `x.array().erf()`?

test/forward.test.cpp
```diff
@@ -5,6 +5,9 @@
 #include <Eigen/Core>
 using namespace Eigen;
 
+// For Special functions like Erf
+#include <unsupported/Eigen/SpecialFunctions>
```
allanleal:
```suggestion
#include <eigen3/unsupported/Eigen/SpecialFunctions>
```

autodiff/forward/forward.hpp
```diff
@@ -1487,6 +1493,15 @@ constexpr void apply(Dual<T, G>& self, AbsOp)
     self.grad *= aux / self.val;
 }
 
+template<typename T, typename G>
+constexpr void apply(Dual<T, G>& self, ErfOp)
+{
+    constexpr double pi = 3.1415926535897932384626433832795029;
+    const T aux = self.val;
+    self.val = erf(aux);
+    self.grad *= 2.0 * exp(-aux*aux)/sqrt(pi);
```
allanleal: Let's precompute `sqrt(pi)`. Unfortunately, I think `std::sqrt` is not `constexpr`, so we set `sqrt_pi` instead of `pi`.

volpatto:
> There is an install fail. I think its due to the Eigen "unsupported" header.
> I don't use conda so can't check this at my end easily.

@allanleal maybe would be nice if a CONTRIBUTING file could be handy for whose that want to contribute with `autodiff`. I think that I will create an Issue for it.

autodiff/forward/forward.hpp
ludkinm: Sorry, are you suggesting: 
``constexpr double sqrt_pi = std::sqrt(3.1415926535897932384626433832795029);``

ludkinm:
@allanleal The Eigen implementation of erf is causing problems (and for reverse mode too) 
I have Eigen 3.3.7 (i.e. git checkout 3.3.7 on their [git mirror](https://github.com/eigenteam/eigen-git-mirror)) and now can't compile. 
I get a THIS_TYPE_IS_NOT_SUPPORTED in the EIGEN_STATIC_ASSERT on line 342 column 67 of ``eigen3/unsupported/Eigen/src/SpecialFunctions/SpecialFunctionsImpl.h``

When I first wrote this, I was using the master of Eigen (which has some updates to that file), this is probably why the github builds are failing.

allanleal:
I think we should then not have examples/tests that use eigen + erf until they have a new version with this sorted out. I also tend to use their master branch, but many pin their version. So, I think this is a solution for now.

allanleal:
@ludkinm I did the fixes. Do you want to work more on this PR or can I merge this?

test/forward.test.cpp
allanleal: I checked, and the following  `x.array().erf()` seems valid but causes the same compilation error with Eigen 3.7.

ludkinm:
@allanleal, sorry for the delay. I think its ok in its current state.

test/forward.test.cpp
ludkinm: @allanleal Just spotted: this include is no longer needed since the test using Eigen was removed

PR #83, opened 2019-11-12
Add Error function to reverse mode

ludkinm:

PR #87, opened 2019-11-13
tweak for erf. sqrt_pi literal in reverse. Removed Eigen header from forward

ludkinm:
This PR tidies up from PR #83 and #84.


allanleal:
Thanks @ludkinm !

PR #89, opened 2019-11-30
Implement new autodiff type based on tape representation

supersega:
# Description

This PR introduce new automatic differentiable variable. 

## Type of change

- Create tape based storage for fast lookup.
- Add variable which store their index in tape storage.
- Overload operators and math function.
- Create example with new type.

### DIFF
--- a/autodiff/mixed/algorithms.hpp
+++ b/autodiff/mixed/algorithms.hpp
@@ -1,0 +1,30 @@
+//                  _  _
+//  _   _|_ _  _|o_|__|_
+// (_||_||_(_)(_|| |  |
+//
+// automatic differentiation made easier in C++
+// https://github.com/autodiff/autodiff
+//
+// Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+//
+// Copyright (c) 2018-2019 Serhii Malyshev
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include <autodiff/mixed/algorithms/first_order.hpp>
--- a/autodiff/mixed/core/meta.hpp
+++ b/autodiff/mixed/core/meta.hpp
@@ -1,0 +1,69 @@
+//                  _  _
+//  _   _|_ _  _|o_|__|_
+// (_||_||_(_)(_|| |  |
+//
+// automatic differentiation made easier in C++
+// https://github.com/autodiff/autodiff
+//
+// Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+//
+// Copyright (c) 2018-2019 Serhii Malyshev
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#pragma once
+
+#include <array>
+#include <tuple>
+
+namespace autodiff::meta
+{
+namespace detail
+{
+/// @brief Meta function to generate tuple which contains derivatives array.
+/// @return Tuple which contain array of derivatives weights.
+constexpr auto generate_tuple_of_derivatives_weights(std::index_sequence<0>)
+{
+    return std::tuple(std::array<double, 2>{ });
+}
+
+/// @brief Meta function to generate tuple which contains derivatives arrays.
+/// @return Tuple which contain arrays of derivatives weights.
+template<std::size_t I, std::size_t... N>
+constexpr auto generate_tuple_of_derivatives_weights(std::index_sequence<I, N...>)
+{
+    constexpr auto cnt = sizeof...(N);
+    return std::tuple_cat(generate_tuple_of_derivatives_weights(std::make_index_sequence<cnt>{}), 
+        std::tuple(std::array<double, cnt + 2>{}));
+}
+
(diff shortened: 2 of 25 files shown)


autodiff/mixed/expressions/binary.hpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /// @brief Right expression value.
+    /// @return Constant reference to right expression value.
+    auto right() const -> const double&
+    {
+        return m_right;
+    }
+private:
+    const double& m_left;  ///< Value reference to left part of binary node.
```
allanleal: Shouldn't left and right also be expressions? What happens if we have `(x + y) + (x * y)`?

autodiff/mixed/expressions/binary.hpp
```diff
@@ -0,0 +1,191 @@
...
+    const double& m_right; ///< Value reference to right part of binary node.
+};
+
+/// @brief Binary sum expression.
+/// @tparam N Maximal derivative order of expression.
+template<std::size_t N>
+struct sum_expression final : binary_expression<N>
```
allanleal: To me, *sum* means a summation of numbers (e.g., sum operator, sum symbol in LaTeX, etc.). I think this should be `addition_expression`.

autodiff/mixed/expressions/binary.hpp
```diff
@@ -0,0 +1,191 @@
...
+    virtual void update()
+    {
+        this->value() = base::left() + base::right();
+    }
+};
+
+/// @brief Binary minus expression.
+/// @tparam N Maximal derivative order of expression.
+template<std::size_t N>
+struct minus_expression final : binary_expression<N>
```
allanleal: Here, we should use `subtraction_expression` if the intent is `x - y`. I think `minus_expression` makes sense for the unary operation `-x`.

autodiff/mixed/expressions/binary.hpp
```diff
@@ -0,0 +1,191 @@
...
+    explicit divide_expression(const double& l, const double& r) : 
+        base(l / r, l, r)
+    { 
+        eval_derivatives();
+    }
+
+    /// @brief Update value of divide expression.
+    /// @details Only first order derivatives are need recalculation,
+    /// since other are always zero.
+    virtual void update()
```
allanleal: Wondering here if these update methods can deal with branches (e.g., if conditions)?

autodiff/mixed/expressions/expression.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+#include <array>
+#include <cmath>
+#include <utility>
+
+#include <autodiff/mixed/core/meta.hpp>
+
+namespace autodiff::taperep {
+
+constexpr auto first_order = 1;  ///< Constant to get first order partial derivatives.
```
allanleal: Do we really need these two constants? Early on during this review, I found `partial_derivatives<first_order>()` and my impression was that `first_order` was a type. `partial_derivatives<1>()` would have been clearer. But this is a bit subjective. Maybe we could name `order1st`, `order2nd`, `order3rd`, `order4th`, etc.

examples/CMakeLists.txt
```diff
@@ -1,2 +1,3 @@
 add_subdirectory(forward)
 add_subdirectory(reverse)
+add_subdirectory(mixed)
```
allanleal: What is the meaning of `mixed`? What would propose as alternative names so we can decide between them?

autodiff/mixed/core/tape.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+#include <vector>
+
+#include <autodiff/mixed/core/variable.hpp>
+#include <autodiff/mixed/expressions/parameter.hpp>
+
+namespace autodiff::taperep 
+{
+/// @brief Class which represent tape based storage for expression tree.
+/// @tparam N order of tape.
```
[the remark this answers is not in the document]
supersega: Yes, we can. I can send publication where algorithm is described.

autodiff/mixed/core/variable.hpp
```diff
@@ -0,0 +1,129 @@
...
+template<std::size_t N>
+struct var final
+{
+    using taped_storage_ptr = tape_storage<N>*;
+    using taped_storage_ref = tape_storage<N>&;
+
+    constexpr static std::size_t order = N;
+
+    template<std::size_t M>
+    friend struct tape_storage;
```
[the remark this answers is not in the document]
supersega: Yes, we have private ctor to create var with index and tape. To ensure that we manage allocations correctly.

autodiff/mixed/core/variable.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+    /// @brief Convert var to double.
+    explicit operator double() const
+    {
+        return value();
+    }
+
+    /// @brief Get value for var.
+    /// @return Numerical value of variable.
+    auto value() const -> const double&
```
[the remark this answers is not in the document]
supersega: We can do it in next PR to avoid monsters =)

autodiff/mixed/core/variable.hpp
```diff
@@ -0,0 +1,129 @@
...
+    auto index() const -> std::size_t
+    {
+        return m_index;
+    }
+
+    /// @brief Get Stack which manage allocation of vars.
+    /// @return Pointer to tape.
+    auto tape() const -> taped_storage_ref
+    {
+        return *m_tape;
```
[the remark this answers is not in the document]
supersega: We can throw here, but I’m not sure if we need it. It will create performance issues.

autodiff/mixed/eigen/eigen_defs.hpp
```diff
@@ -0,0 +1,90 @@
...
+// SOFTWARE.
+
+#pragma once
+
+namespace Eigen {
+
+template<typename T>
+struct NumTraits;
+
+template<> struct NumTraits<autodiff::taperep::var<1>> : NumTraits<double> // permits to get the epsilon, dummy_precision, lowest, highest functions
```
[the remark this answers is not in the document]
supersega: Probably yes, I will try.

autodiff/mixed/expressions/binary.hpp
supersega: No, it should not. We store reference to parent values. All expressions have value and they don’t change.

autodiff/mixed/expressions/binary.hpp
supersega: Also can be introduced in new PR

autodiff/mixed/expressions/expression.hpp
supersega: I like first_order. I can read it like phrase from book.

autodiff/mixed/core/meta.hpp
```diff
@@ -0,0 +1,69 @@
...
+#pragma once
+
+#include <array>
+#include <tuple>
+
+namespace autodiff::meta
+{
+namespace detail
+{
+/// @brief Meta function to generate tuple which contains derivatives array.
```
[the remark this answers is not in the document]
supersega: We should just arrange about documentation style, and I will do ;)

autodiff/mixed/algorithms.hpp
```diff
@@ -0,0 +1,30 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include <autodiff/mixed/algorithms/first_order.hpp>
```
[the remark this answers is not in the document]
supersega: I believe that `_` will help to reed file name, but if we prefer such naming as you mentioned I will change.

autodiff/mixed/core/variable.hpp
```diff
@@ -0,0 +1,129 @@
...
+    template<std::size_t M>
+    friend struct tape_storage;
+
+    /// @brief Default constructor is default.
+    var() = default;
+
+    /// @brief Construct variable from arithmetic types. We need it because Eigen requires it.
+    /// @details In Eigen code they have check in if statement, where compiler
+    /// looking for Scalar(0). Our constructor is empty.
+    template<typename U, std::enable_if_t<std::is_arithmetic_v<U>>...>
```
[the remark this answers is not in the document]
supersega: Yes, we must support custom floating point numbers, but we can do it in someone of future PRs.

examples/CMakeLists.txt
supersega: We can travel in both directions.

supersega:
Hi @allanleal, I want to replace only most serious moments like naming style etc.
#### What I suggest to move into future PRs?
- Add examples.
- Make enable to work with custom numbers. (Probably it will be common check for all autodiff types)
- Introduce high order derivatives.
- Documentation.
#### What I suggest to fix in this PR?
- File naming.
- Java style for Doxygen.
- Add some tests.

PaulXiCao:
@allanleal There has not been any work on this mr for quite some time. Is this still a wanted feature? If so how can I help? Should I create a copy of this mr and improve on it? Is there a way to split it into smaller chunks? Maybe a milestone with multiple smaller steps in-between which could be iteratively merged?

ianhbell:
Could you comment on the potential advantages of this new type? I am in particular interested in obtaining all the second and third partial derivatives of a function of two variables in one shot, and wonder whether this would be superior to multiple calls with forward differentiation.

allanleal:
Ian, let's talk about this (please email me). I think I have a better solution for you (which we could test first before releasing). I assume your use case is for thermodynamics; and the two variables are perhaps temperature and density?!

ianhbell:
Correct - email headed your way

PR #101, opened 2020-02-05
Using conda package instead of pip

marcelotrevisani:


environment.devenv.yml
```diff
@@ -8,9 +8,8 @@ dependencies:
   - gxx_linux-64=7.3.0  # [linux]
   - ccache  # [unix]
   - pip
-  - pip:
-    - mkdocs
-    - mkdocs-material
-    - mkdocs-macros-plugin
+  - mkdocs
+  - mkdocs-material
+  - mkdocs-macros-plugin
```
volpatto: Oh, great! I have been thinking about get MkDocs from `conda-forge`. Actually a colleague pointed it to me last week, I think.

marcelotrevisani: It was missing ``mkdocs-macros-plugin`` and another package which was a dependency of it. So, I just added those to the ``conda-forge`` channel

PR #116, opened 2020-06-01
Add atan2

pettni:
Really enjoy this library, thanks for building it.

I needed ```atan2``` to use ```autodiff``` together with [```Sophus```](https://github.com/strasdat/Sophus) for gradient-based optimization on Lie Groups. I thought it might be a good idea to add it by default.

The forward version is a bit verbose but could be shortened at the cost of more memory allocation. It seemed to me like this way was more in line with the rest of the library, but since ```atan2``` is relatively obscure compared to other binary operations it could make sense to prioritize code simplicity.

### DIFF
--- a/autodiff/forward/forward.hpp
+++ b/autodiff/forward/forward.hpp
@@ -49,6 +49,7 @@ using std::abs;
 using std::acos;
 using std::asin;
 using std::atan;
+using std::atan2;
 using std::cos;
 using std::exp;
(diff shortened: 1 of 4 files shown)


allanleal:
Hi Peter. Wow, I'm glad you figured out by yourself how to extend the lib with `atan2`! 
All looks good. Thanks a lot for this.

PR #123, opened 2020-08-25
Add std::hypot to forward and reverse mode

benjaminhuth:
Hi,

For a project I'm currently working on I needed support for `std::hypot`, and I have added support for this both for the binary- and the ternary overload in both the forward and the reverse mode. The tests are also updated. It would be great if these changes could be merged in.


allanleal (approved):
Thanks for this (and sorry for the delay) - it looks good! I hope you can manage any issue arising from the use of these functions! ;)

PR #125, opened 2020-09-15
resolve some warnings (-Wextra, -Wshadow, -Wnon-virtual-dtor)

benjaminhuth:
Hi,

this is not actually very important, but at least for the project I'm working on these warnings trigger, so I decided to resolve them. Maybe there is also some benefit for other people... 
There shouldn't be any change in functionality, just some variable renaming and additional default constructors/destructors...). I also did not resolve the warnings in the tests. The only warnings I still get are related to Eigen.

### DIFF
--- a/autodiff/reverse/reverse.hpp
+++ b/autodiff/reverse/reverse.hpp
@@ -243,12 +243,15 @@ template<typename T, typename U, EnableIf<isArithmetic<U>>...> bool operator>(co
 /// The abstract type of any node type in the expression tree.
 template<typename T>
 struct Expr
-{
+{    
     /// The value of this expression node.
     T val = {};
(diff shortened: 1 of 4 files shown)


autodiff/reverse/reverse.hpp
```diff
@@ -243,12 +243,15 @@ template<typename T, typename U, EnableIf<isArithmetic<U>>...> bool operator>(co
 /// The abstract type of any node type in the expression tree.
 template<typename T>
 struct Expr
-{
+{
```
allanleal, line 246:
```suggestion
{
```

allanleal (approved):
Thanks, this is indeed helpful.

PR #126, opened 2020-09-26
Integrate bazel.build 

patrickhart:
# Description
This PR introduces [bazel.build](https://bazel.build/) for the usage within bazel projects

# Type of change
- Bazel WORKSPACE file
- Bazel BUILD files
- Integration of examples as cc_test

# Benefits
- Could enable a better CI testing
- People can use autodiff in their Bazel projects easily

### DIFF
--- a/examples/forward/BUILD
+++ b/examples/forward/BUILD
@@ -1,0 +1,100 @@
+cc_test(
+    name = "example-forward-single-variable-function",
+    srcs=["example-forward-single-variable-function.cpp"],
+    deps = ["//:autodiff"],
+    copts = ["-I./"]
+)
+
+cc_test(
+    name = "example-forward-single-variable-function-custom-scalar",
+    srcs = ["example-forward-single-variable-function-custom-scalar.cpp"],
+    deps = ["//:autodiff"],
+    copts = ["-I./"]
+)
+
+cc_test(
+    name = "example-forward-multi-variable-function",
+    srcs=["example-forward-multi-variable-function.cpp"],
+    deps = ["//:autodiff"],
+    copts = ["-I./"]
+)
+
+cc_test(
+    name = "example-forward-multi-variable-function-with-parameters",
+    srcs = ["example-forward-multi-variable-function-with-parameters.cpp"],
+    deps = ["//:autodiff"],
+    copts = ["-I./"]
+)
+
+cc_test(
+    name = "example-forward-jacobian-derivatives-using-eigen",
+    srcs = ["example-forward-jacobian-derivatives-using-eigen.cpp"],
+    deps = ["//:autodiff"],
+    copts = ["-I./"]
+)
+
+cc_test(
+    name = "example-forward-jacobian-derivatives-using-eigen-with-scalar",
+    srcs = ["example-forward-jacobian-derivatives-using-eigen-with-scalar.cpp"],
+    deps = ["//:autodiff"],
+    copts = ["-I./"]
+)
+
+cc_test(
+    name = "example-forward-jacobian-derivatives-using-eigen-with-parameters",
+    srcs = ["example-forward-jacobian-derivatives-using-eigen-with-parameters.cpp"],
+    deps = ["//:autodiff"],
+    copts = ["-I./"]
+)
+
+cc_test(
+    name = "example-forward-higher-order-derivatives",
+    srcs = ["example-forward-higher-order-derivatives.cpp"],
+    deps = ["//:autodiff"],
+    copts = ["-I./"]
+)
+
+cc_test(
+    name = "example-forward-hessian-derivatives-using-eigen",
+    srcs = ["example-forward-hessian-derivatives-using-eigen.cpp"],
+    deps = ["//:autodiff"],
+    copts = ["-I./"]
+)
+
+cc_test(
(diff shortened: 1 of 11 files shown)


allanleal:
Hi, this is quite interesting - thanks. I'll start a review now. 

The CI is failing because of the added `autodiff/BUILD` file and we are using `autodiff/build` as the directory where the tests/examples are built.

examples/forward/BUILD
```diff
@@ -0,0 +1,100 @@
+cc_test(
```
allanleal, line 1: Is there an automatic approach that will work any time a new example file is added? This file may get forgotten when the list of examples gets expanded over time.

examples/reverse/BUILD
```diff
@@ -0,0 +1,48 @@
+cc_test(
```
allanleal, line 1: The same here.

allanleal:
To fix the CI errors, could you please change `build` to `.build` in the files:

https://github.com/autodiff/autodiff/blob/master/ci/actions/install.sh
https://github.com/patrickhart/autodiff/blob/master/.github/workflows/windows.yml

This will fix the name conflict of `BUILD` file and `build` dir.

Thanks a lot.

examples/forward/BUILD
patrickhart: I am not aware of such a functionality. Will have a look if it is possible to generate/group these automatically.

patrickhart:
> To fix the CI errors, could you please change `build` to `.build` in the files:
> 
> https://github.com/autodiff/autodiff/blob/master/ci/actions/install.sh
> https://github.com/patrickhart/autodiff/blob/master/.github/workflows/windows.yml
> 
> This will fix the name conflict of `BUILD` file and `build` dir.
> 
> Thanks a lot.

Will have a look here as well.

patrickhart:
So just to add some "documentation", autodiff could be used in a Bazel project then as follows:


WORKSPACE file
```
workspace(name = "optimization_research")
load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

git_repository(
    name = "lib_autodiff",
    branch = "master",
    remote = "https://github.com/patrickhart/autodiff",
)

http_archive(
    name = "com_github_eigen_eigen",
    build_file_content = """
cc_library(
    name = 'eigen',
    srcs = [],
    includes = ['.'],
    hdrs = glob(['Eigen/**']),
    visibility = ['//visibility:public'],
)
""",
    sha256 = "4b1120abc5d4a63620a886dcc5d7a7a27bf5b048c8c74ac57521dd27845b1d9f",
    strip_prefix = "eigen-git-mirror-98e54de5e25aefc6b984c168fb3009868a93e217",
    urls = ["https://github.com/eigenteam/eigen-git-mirror/archive/98e54de5e25aefc6b984c168fb3009868a93e217.zip"]
)
```

And BUILD file
```
cc_binary(
    name = "optimize",
    srcs = ["optimize.cpp"],
    deps = ["@lib_autodiff//:autodiff"],
    copts = ["-I./external/lib_autodiff/"],
    visibility = ["//visibility:public"],
)
```

And then all examples provided in the autodiff can be used in the derived Bazel project.

As a side note: `bazel test //...` runs all provided examples, which could as be handy for the CI.

allanleal:
OK, CI looks good now! Just check please if an automated way of building `*.cpp` example files is possible with bazel (similar to CMake glob functionality). If not possible, then we go ahead.

patrickhart:
> OK, CI looks good now! Just check please if an automated way of building `*.cpp` example files is possible with bazel (similar to CMake glob functionality). If not possible, then we go ahead.

Using
```
cc_test(
    name = "forward_examples",
    srcs = glob(["*.cpp"]),
    deps = ["//:autodiff"],
    copts = ["-I./"]
)
```

it is possible to collect all .cpp files.

However, as all tests have a main function declared this would lead to duplicate symbols, e.g.:

```
duplicate symbol '_main' in:
    bazel-out/darwin-fastbuild/bin/examples/forward/_objs/forward_examples/example-forward-gradient-derivatives-using-eigen-with-parameters.pic.o
    bazel-out/darwin-fastbuild/bin/examples/forward/_objs/forward_examples/example-forward-single-variable-function.pic.o
ld: 16 duplicate symbols for architecture x86_64
```

So for now I do not see a better option to integrate an automated *.cpp collection, besides converting the examples into e.g. gtests. Also, if not declared separately it would not be possible to execute a single, specific example, using e.g. `bazel run //examples/forward:example-forward-gradient-derivatives-using-eigen-with-parameters`.

allanleal:
With CMake, we can collect all example files and generate an executable for each one. I was hoping something similar exists with baze.

patrickhart:
I think that is not (out of the box) possible.. Bazel requires a declaration for a cc_test or cc_binary.
Otherwise it would not be possible to run a specific target using `bazel run //autodiff/examples:xxx`.

Bazel certainly isn't a perfect build tool, but it can provide benefits for CIs/sandboxing/or when working with multiple repositories. As the CI uses CMake, I guess it would not really be required to achieve a 100% coverage of the examples using Bazel.

allanleal:
OK, I'll merge this now. Thanks for this contribution.

PR #130, opened 2020-10-17
Return zero sub-gradient for abs(x) at x=0

pettni:
Perhaps a minor issue, but there is currently a discrepancy between the reverse and forward modes for ```abs(x)``` at ```x=0```, where reverse mode returns derivative ```+1``` and forward results in a NaN after division by zero. Though the regular derivative is undefined at ```x=0``` it seems better to compute one of the sub-derivatives in ```[-1, 1]``` instead of dividing by zero.

I chose 0 here which seemed nice and symmetric and is what e.g. ```pytorch``` does.


allanleal:
Excellent. Thanks for reporting this issue and providing a fix!

PR #132, opened 2020-11-02
[WIP] Improvements for dev mode

volpatto:
**WIP**

I have been working towards improvements for `autodiff` development. I am willing to try some Python bindings tests in the future, so I think that some changes can be useful.

Of course, despite the words "improve"/"improvements" used here, this PR not necessarily "improves" things. We can discuss everything :)


volpatto:
It looks like that autodiff's `setup-miniconda` action is "deprecated" and old. A new version is [here](https://github.com/conda-incubator/setup-miniconda). This old version is breaking `master` as well. I will try to fix it in this PR.

volpatto:
> It looks like that autodiff's `setup-miniconda` action is "deprecated" and old, breaking Windows build. A new version is available [here](https://github.com/conda-incubator/setup-miniconda). This old version is breaking Windows `master` as well. I will try to fix it in this PR.

Fixed.

ci/actions/install.sh
```diff
@@ -7,7 +7,4 @@ conda update -q conda
 conda info -a
 conda devenv
 source activate autodiff
-mkdir .build
-cd .build || exit
-cmake .. -GNinja
-ninja
+inv compile -c -n 2
```
allanleal, line 10: Hi @volpatto , all looks good! Thanks a lot for this. I would prefer, however, to have a simpler solution for compiling and running tests in Windows. The `tasks.py` file below would have to be dropped (I did not understand why it is necessary - it is quite long in comparison with the ninja commands above, and it would be great if we keep only CMake to managing the build tasks for the CI). Do we really need `tasks.py`?

allanleal (commented):
As to the Python wrappers, I already have them in my personal branch for the next major version of `autodiff`. This uses `pybind11` and also includes support to Eigen vectors/matrices with autodiff number types.

ci/actions/install.sh
volpatto: Thanks, @allanleal! I will tell you my opinion on each point.

IMHO, I think that this is why `invoke` is an interesting tool. We have the same command for each platform, the developer does not have to be concerned about several commands to use a build system. From the developer's point of view, compilation is `inv compile` and running the tests is `inv tests`. He/she does not have to be concerned about where is the test executable, neither the options to a pretty display of tests in the terminal. `inv` is a way to turn the develop mode the same for all platforms. So for me, it is simpler this way, because you hide the details of build system commands and use the same despite the OS. But, of course, this is my opinion and you are the core developer (and main developer). If there is no point to you, it's OK! We drop it, no worries.

> I did not understand why it is necessary - it is quite long in comparison with the ninja commands above, and it would be great if we keep only CMake to managing the build tasks for the CI

Actually, it is not _necessary_, but I think it can help. You can use `inv compile` once (say, in the first time you create your fork), to create the build for you, and then use `ninja` thereafter, no problems. Or you can do as implemented before, with commands to create a dir, to generate files with CMake and build. Again, I prefer the way to use `inv compile` encapsulating everything than running more commands, sounds more error-prone to me. But, again, if it's not useful for your workflow, I think that I should drop it, and there is no problem for me :) And BTW, CMake is managing the build tasks, `inv` encapsulates the CMake commands: for instance, `inv compile -c -n 4` is a clean (c) CMake build and install using 2 processes (n).

About everything: my vision is biased because I'm used to this workflow. I'm an occasional `autodiff` dev, you are a frequent dev, so please consider if these changes here are useful for you. I will comment below about what each task below.

tasks.py
```diff
@@ -0,0 +1,206 @@
+from invoke import task
```
volpatto, line 1: Despite the size of this file, only 4 tasks are defined here. I will leave a brief description of each task.

tasks.py
```diff
@@ -0,0 +1,206 @@
...
+            build_subdirectory=BUILD_DIR_DEFAULT / "msvc",
+        )
+        cmake_command = _get_cmake_command(build_dir=build_dir, cmake_generator="Visual Studio 15 2017",
+                                           cmake_arch="x64", config=config)
+        os.chdir(build_dir)
+        c.run(cmake_command)
+
+
+@task
+def compile(c, clean=False, config='Release', number_of_jobs=-1, verbose=False, gen_wrappers=False):
```
volpatto, line 131: This is the compile task. It builds and installs `autodiff`. All targets are considered.

Usage:
```bash
inv compile (options)
```

You can set some build configs with the following options:

* `-c` or `--clean`: set if a build is clean or not. A clean build removes the build dir before perform the build process. Default is `False`
* `--config`. Set if the build is Release of Debug mode. Default is `Release`
* `n` or `--number-of-jobs`. Number of processes to build and install `autodiff`. Default is `-1`, which is the serial option
* `-v` or `--verbose`. Set if the build and install task is verbose. Default is `False`
* `-g` or `--gen-wrappers`. Set if `conda` binaries should be wrapped (it means: symbolic links) to a dir inside build directory. I will give details about this task in another comment. Default is `False`

tasks.py
```diff
@@ -0,0 +1,206 @@
...
+                'Error: Commands to configure MSVC environment variables not found.',
+                code=1,
+            )
+
+    os.chdir(build_dir)
+    c.run("&&".join(commands))
+
+
+@task
+def clear(c, build_dir_path=BUILD_DIR_DEFAULT):
```
volpatto, line 177: A task to remove the build directory. If the user used a non-default build dir, he/she can inform where the build dir is.

Usage:
```bash
inv clear (options)
```

Options:

* `-b` or `--build-dir-path (path to build dir)`. Remove the informed build dir.

tasks.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    Clear autodiff build directory
+    """
+    remove_directory(build_dir_path)
+
+
+@task
+def wrappers(c, wrappers_dir=BUILD_DIR_DEFAULT / "wrappers/conda"):
```
volpatto, line 185: The `wrappers` task is useful because it creates symlinks to conda binaries. So instead of search for binaries in the `miniconda` installation dir in your file system, this task wraps everything for you to a given directory. This directory is `(BUILD_DIR)/wrappers/conda` by default. It is useful, at least for me, because helps me to configure my IDE (CLion, pointing which cmake to use, which python, compilers and etc. The same is valid for VS Code).

Usage:
```bash
inv wrappers (options)
```

Options:
* `-w` or `--wrappers-dir`. Specify the dir you would like to create the wrappers (symlinks). Default is the default build dir

tasks.py
```diff
@@ -0,0 +1,206 @@
...
+    else:
+        print(f"Generating conda wrappers to {wrappers_dir} from {os.environ['CONDA_PREFIX']}/bin")
+
+    generate_wrappers_command = _get_wrappers_command(wrappers_dir)
+    echo(c, generate_wrappers_command)
+    c.run(generate_wrappers_command, pty=True, warn=True)
+
+
+@task
+def tests(c):
```
volpatto, line 201: The `tests` task runs `autodiff`'s `Catch2` tests with a "pretty" output, close to `pytest`. It is simply the following command behind the scenes: `{test_executable} -s -r compact`. Please note that there is no need to know where is the `{test_executable}`, it is encapsulated and configure inside the task.

Usage:
```
inv tests
```

No options are available at this point.

volpatto:
> As to the Python wrappers, I already have them in my personal branch for the next major version of `autodiff`. This uses `pybind11` and also includes support to Eigen vectors/matrices with autodiff number types.

Oh, great, then! I will do it in my personal brach for "exercise" purposes only, as you already have it ready for the future.

volpatto:
I don't know why Python in Windows can't detect `AUTODIFF_BUILD_DIR` environment variable define in `environment.devenv.yml`. I have tried the best I can, this is a pretty basic error though.

volpatto:
As we have discussed privately, I will close this PR and open different PRs for each change. The priority is to fix Windows CI.

PR #133, opened 2020-11-10
Fix Win CI updating setup-miniconda action

volpatto:
Self-described.


allanleal:
Super! Thanks, @volpatto !

PR #135, opened 2020-11-18
Update setup-miniconda GH action version

volpatto:
GH Actions had an update in October that require us to update `setup-miniconda` in `autodiff` CI. The `setup-miniconda` needs to be bumped to `v2` for all builds.

I'm updating it because one day the CI build would break.


allanleal:
Thanks a lot, @volpatto !

PR #136, opened 2020-11-18
resolving -Wunused-parameter warnings for clang

benjaminhuth:
This PR does resolves some warnings triggered by `-Wunused-parameter` on clang. It would be great if this could be merged in, since this triggers the MacOS CI configuration in a project where I want to use `autodiff`.


allanleal:
Thanks!

PR #155, opened 2021-04-27
Fixes for higher order dual numbers

supersega:
# Make forward mode more compatible with HigherOrderDual numbers

1. Allow cast vectors of HigherOrderDual to Eigen::VectorXd
2. Fix problem with expression eval for HigherOrderDual related to wrong recursive ValueType meta function.
3. Make Eigen related functions compatible with HigherOrderDual.

### DIFF
--- a/test/forward.test.cpp
+++ b/test/forward.test.cpp
@@ -1079,4 +1079,135 @@ TEST_CASE("VectorXdual2nd tests", "[dual2nd]")
         for(auto i = 0; i < 3; ++i)
             REQUIRE( x(i) == approx(y(i)) );
     }
+
+    using dual2nd = HigherOrderDual<2>;
+
+    SECTION("testing gradient derivatives")
+    {
+        // Testing complex function involving sin and cos
+        auto f = [](const VectorXdual2nd& x) -> dual2nd
+        {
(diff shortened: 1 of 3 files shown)


supersega:
@allanleal could you please have a look?

test/forward.test.cpp
```diff
@@ -1059,3 +1066,148 @@ TEST_CASE("Eigen::VectorXdual tests", "[dual]")
...
+{
+    using dual3rd = HigherOrderDual<2>;
+    using VectorXdual3rd = Eigen::Matrix<dual3rd, -1, 1, 0, -1, 1>;
+
+    SECTION("testing casting to VectorXd")
+    {
+        VectorXdual3rd x(3);
+        x << 1.0, 2.0, 3.0;
+        
+        VectorXd y = x.template cast<double>();
```
allanleal, line 1152: Interesting, `x.cast<double>()` does not compile?

allanleal (approved):
Looks good. Thanks for the fixes!

test/forward.test.cpp
supersega: Yes, because for dual2nd T is dual<double, double > and this leads to cast only for Dual<double, double >

PR #171, opened 2021-07-26
Modified build files path in .gitignore

YuanweiZHANG:
I need to use autodiff for a C++ project. When I try to install it by CMake, I found that files under the .build directory haven't be ignored yet.


allanleal:
Thanks!

PR #176, opened 2021-08-12
Implement reverse prop condition expr

### LINKED ISSUE #88: Implement ConditionExpr for reverse mode algorithm
@ludkinm @supersega

This is a follow up of [this discussion](https://github.com/autodiff/autodiff/issues/40#issuecomment-554226543_). 

Consider the following function:

~~~c++
var f(var x) { return (x > 0) ? x**x : x*x*x; } 
~~~

Consider now what happens in the code below:

~~~c++
var x, u;

x = 1.0; 
u = f(x); // expression tree constructed for x*x

x = -1.0; 
u = f(x); // expression tree constructed for x*x*x
~~~

Each function call is re-constructing the expression tree. To be able to avoid this, we need a `ConditionExpr` and `condition` function, so that future evaluations of `u` could be done by just changing `x`, and then applying an update request on `u`, for example:

~~~c++
var f(var x) { return condition(x > 0, x**x, x*x*x); } 
~~~

which would permit us to do:

~~~c++
x = 1.0; 
u = f(x);

x = -1.0; 
u.update(); // this functionality also needs to be implemented
~~~

I think with this addition, will truly be able to no longer require expression tree reconstruction, which has happened so far because of potential branching in expressions involving `var` numbers.

_Originally posted by @allanleal in https://github.com/autodiff/autodiff/issues/40#issuecomment-554226543_

jargonzombies:
Initial implementation only, only supports simple boolean expressions
comparing Expr<T> and related classes, not yet conjunctions between
boolean expressions themselves

- Add nullary update fn to Expr<T> interface for back-propagating value
  changes in nodes of an expression tree, implement for all existing
  derived classes
- Replace virtual keyword with override in Expr<T>-derived classes
- Introduce BooleanExpr to capture arbitrary comparison between two
  Expr<T>
- Introduce ConditionalExpr to capture switching between two expressions
  depending on the value of a BooleanExpr
- Change boolean operators to return BooleanExpr's
- Alter semantics of Variable<T>::operator=(Numeric ) and
  Variable<T>::operator[+-*/]=(Numeric x) to alter the underlying
  independent variable if set without introducing a new independent
  variable expression
- Add test for condition expression
- Fix .clang-format and reformat var.hpp

Closes #88

### DIFF
--- a/autodiff/reverse/var/var.hpp
+++ b/autodiff/reverse/var/var.hpp
@@ -36,6 +36,7 @@
 #include <cstddef>
 #include <functional>
 #include <memory>
+#include <stdexcept>
 
 // autodiff includes
 #include <autodiff/common/meta.hpp>
@@ -134,7 +135,6 @@ struct isVariable { constexpr static bool value = false; };
 
 template<typename T>
 struct isVariable<Variable<T>> { constexpr static bool value = true; };
-
 } // namespace traits
 
 template<typename T>
@@ -1187,9 +1187,11 @@ struct ConditionalExpr : Expr<T>
     using Expr<T>::val;
     ExprPtr<T> l, r;
 
+    ConditionalExpr(const std::shared_ptr<BooleanExpr>& wrappedPred, const ExprPtr<T>& ll, const ExprPtr<T>& rr) : Expr<T>(*wrappedPred ? ll->val : rr->val), predicate(wrappedPred), l(ll), r(rr) {}
+
     template<typename Derived, EnableIf<std::is_base_of_v<BooleanExpr, Derived>>...>
-    ConditionalExpr(const Derived& pred, const ExprPtr<T>& ll, const ExprPtr<T>& rr)
-        : Expr<T>(pred ? ll->val : rr->val), predicate(pred.wrap()), l(ll), r(rr) {}
+    ConditionalExpr(const Derived& pred, const ExprPtr<T>& left, const ExprPtr<T>& right)
+        : ConditionalExpr(pred.wrap(), left, right) {}
 
     void propagate(const T& wprime) override
     {
@@ -1199,8 +1201,8 @@ struct ConditionalExpr : Expr<T>
 
     void propagatex(const ExprPtr<T>& wprime) override
     {
-        if(predicate->val) l->propagatex(wprime);
-        else r->propagatex(wprime);
+        l->propagatex(derive(wprime, constant<T>(0.0)));
+        r->propagatex(derive(constant<T>(0.0), wprime));
     }
 
     void update() override
@@ -1213,6 +1215,10 @@ struct ConditionalExpr : Expr<T>
             r->update();
             this->val = r->val;
         }
+    }
+
+    ExprPtr<T> derive(const ExprPtr<T>& left, const ExprPtr<T>& right) const {
+      return std::make_shared<ConditionalExpr>(*predicate, left, right);
     }
 };
 
@@ -1332,35 +1338,24 @@ struct Variable
     /// Update the value of this variable with changes in its expression tree
     void update() { expr->update(); }
 
+    void update(T value) {
+      if(auto independentExpr = std::dynamic_pointer_cast<IndependentVariableExpr<T>>(expr)) {
+        independentExpr->val = value;
+        independentExpr->update();
+      } else {
+        throw std::logic_error("Cannot update the value of a dependent expression stored in a variable");
+      }
+    }
+
     /// Implicitly convert this Variable object into an expression pointer.
     operator const ExprPtr<T>&() const { return expr; }
 
     /// Explicitly convert this Variable object into its underlying arithmetic type.
     explicit operator T() const { return expr->val; }
 
-    struct Overwrite { T operator()(const T& /* ignore */, const T& val) { return val; } };
-
-    template<typename U, class Op, EnableIf<isArithmetic<U>>...>
-    bool tryIndependentUpdate(const U& val, Op&& op = Op{})
-    {
-        if(auto independentExpr = std::dynamic_pointer_cast<IndependentVariableExpr<T>>(expr)) {
-            independentExpr->val = op(independentExpr->val, val);
-            independentExpr->update();
-            return true;
-        }
-
-        return false;
-    }
-
     /// Assign an arithmetic value to this variable.
     template<typename U, EnableIf<isArithmetic<U>>...>
-    auto operator=(const U& val) -> Variable&
-    {
-        if(!tryIndependentUpdate(val, Overwrite{})) {
-            *this = Variable(val);
-        }
-        return *this;
-    }
+    auto operator=(const U& val) -> Variable& { *this = Variable(val); return *this; }
 
     /// Assign an expression to this variable.
     auto operator=(const ExprPtr<T>& x) -> Variable& { *this = Variable(x); return *this; }
@@ -1371,19 +1366,11 @@ struct Variable
     Variable& operator*=(const ExprPtr<T>& x) { *this = Variable(expr * x); return *this; }
     Variable& operator/=(const ExprPtr<T>& x) { *this = Variable(expr / x); return *this; }
 
-    // Assignment operators with arithmetic values
-    template<typename U, class Op>
-    Variable& arithmeticAlteration(const U& x, Op&& op)
-    {
-        if(!tryIndependentUpdate(x, op)) {
-            *this = Variable(op(expr, x));
-        }
-        return *this;
-    }
-    template<typename U, EnableIf<isArithmetic<U>>...> Variable& operator+=(const U& x) { return arithmeticAlteration(x, std::plus<>{}); }
-    template<typename U, EnableIf<isArithmetic<U>>...> Variable& operator-=(const U& x) { return arithmeticAlteration(x, std::minus<>{}); }
-    template<typename U, EnableIf<isArithmetic<U>>...> Variable& operator*=(const U& x) { return arithmeticAlteration(x, std::multiplies<>{}); }
-    template<typename U, EnableIf<isArithmetic<U>>...> Variable& operator/=(const U& x) { return arithmeticAlteration(x, std::divides<>{}); }
+	// Assignment operators with arithmetic values
+    template<typename U, EnableIf<isArithmetic<U>>...> Variable& operator+=(const U& x) { *this = Variable(expr + x); return *this; }
+    template<typename U, EnableIf<isArithmetic<U>>...> Variable& operator-=(const U& x) { *this = Variable(expr - x); return *this; }
+    template<typename U, EnableIf<isArithmetic<U>>...> Variable& operator*=(const U& x) { *this = Variable(expr * x); return *this; }
+    template<typename U, EnableIf<isArithmetic<U>>...> Variable& operator/=(const U& x) { *this = Variable(expr / x); return *this; }
 };
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


jargonzombies:
This isn't done just yet but it's probably worth taking a look to see if you agree with the direction @allanleal.

Missing to my eyes:
- [ ] More docstrings
- [ ] `BooleanExpr` logical conjunctions, e.g. `condition(2 < x && x < 4, x * x, - x)`

autodiff/reverse/var/var.hpp
```diff
@@ -67,72 +69,127 @@ using std::sinh;
 using std::sqrt;
 using std::tan;
 using std::tanh;
-using std::hypot;
 
-template<typename T> struct Expr;
```
allanleal, line 72: Please keep these declarations on a single line. It's much more readable IMO the previous way.

autodiff/reverse/var/var.hpp
```diff
@@ -154,92 +211,149 @@ ExprPtr<T> constant(const T& val);
 //------------------------------------------------------------------------------
 // ARITHMETIC OPERATORS (DECLARATION ONLY)
 //------------------------------------------------------------------------------
-template<typename T> ExprPtr<T> operator+(const ExprPtr<T>& r);
```
allanleal, line 157: Here too - keep on same line (otherwise it gets very noisy).

allanleal: These changes are also making it difficult for me to recognize any possible new addition among the formatting changes.

autodiff/reverse/var/var.hpp
```diff
@@ -902,160 +1224,535 @@ struct Hypot3Expr : TernaryExpr<T>
...
+ExprPtr<T> constant(const T& val)
+{
+    return std::make_shared<ConstantExpr<T>>(val);
+}
 
 //------------------------------------------------------------------------------
-// COMPARISON OPERATORS
+// ARITHMETIC OPERATORS
 //------------------------------------------------------------------------------
-template<typename T> bool operator==(const ExprPtr<T>& l, const ExprPtr<T>& r) { return l->val == r->val; }
```
allanleal, line 1016: Here as well - it's more readable if on the same line.

allanleal (commented):
Hi Jan, this is looking really good. Thanks a lot for your initiative in implementing this feature. Seems like we are both at ETH Zurich! :)

autodiff/reverse/var/var.hpp
```diff
@@ -911,19 +1058,168 @@ struct Hypot3Expr : TernaryExpr<T>
...
+      }
+    };
+
+    bool val = {};
+
+    explicit BooleanExpr(bool v) : val(v) {}
+    operator bool() const { return val; }
+
+    virtual void update() = 0;
+    virtual std::shared_ptr<BooleanExpr> wrap() const = 0;
```
allanleal, line 1101: What's the `wrap` method for?

autodiff/reverse/var/var.hpp
```diff
@@ -1091,42 +1374,78 @@ struct Variable
...
+using expr_common_t = std::common_type_t<decltype(ExprValue::value(std::declval<T>())), decltype(ExprValue::value(std::declval<U>()))>;
+
+template<class> struct sfinae_true : std::true_type {};
+template<typename T> static auto is_expr_test(int) -> sfinae_true<decltype(ExprValue::value(std::declval<T>()))>;
+template<typename T> static auto is_expr_test(long) -> std::false_type;
+template<typename T> struct is_expr : decltype(is_expr_test<T>(0)) {};
+template<typename T> constexpr bool is_expr_v = is_expr<T>::value;
+
+template<typename T>
+struct CoerceExpr {
```
allanleal, line 1396: These `coerce` methods can be free (template) functions (instead of static methods of `CoerceExpr`). The name `CoerceExpr` can also confuse, because `SomethingExpr` in `var.hpp` is usually a new type derived from `Expr`.

autodiff/reverse/var/var.hpp
```diff
@@ -1091,42 +1374,78 @@ struct Variable
...
-
-template<typename T, typename U, EnableIf<isArithmetic<U>>...> bool operator==(const Variable<T>& l, const U& r) { return l.expr == r; }
-template<typename T, typename U, EnableIf<isArithmetic<U>>...> bool operator!=(const Variable<T>& l, const U& r) { return l.expr != r; }
-template<typename T, typename U, EnableIf<isArithmetic<U>>...> bool operator<=(const Variable<T>& l, const U& r) { return l.expr <= r; }
-template<typename T, typename U, EnableIf<isArithmetic<U>>...> bool operator>=(const Variable<T>& l, const U& r) { return l.expr >= r; }
-template<typename T, typename U, EnableIf<isArithmetic<U>>...> bool operator<(const Variable<T>& l, const U& r) { return l.expr < r; }
-template<typename T, typename U, EnableIf<isArithmetic<U>>...> bool operator>(const Variable<T>& l, const U& r) { return l.expr > r; }
+
+struct ExprValue {
+  template<typename T, EnableIf<isArithmetic<T>>...> static auto value(const T& t) { return t; }
```
allanleal, line 1381: Similar to my comments below for `coerce`, we can have these `value` methods as free functions.

autodiff/reverse/var/var.hpp
```diff
@@ -911,19 +1058,168 @@ struct Hypot3Expr : TernaryExpr<T>
...
+    explicit BooleanExpr(bool v) : val(v) {}
+    operator bool() const { return val; }
+
+    virtual void update() = 0;
+    virtual std::shared_ptr<BooleanExpr> wrap() const = 0;
+};
+
+/// Capture numeric comparison between two expression trees
+template<typename T, typename Comparator>
+struct ExpressionComparison : public BooleanExpr
```
allanleal, line 1106: Wondering if we could eliminate this type (`ExpressionComparison`) by using a lambda function with mutable capture. 

The lambda function would be created in the body of the comparison operators (`==`, `!=`, `<`, `>=`, etc.). Something like this (not exactly like this!):

~~~c++
template<typename T>
auto operator==(const ExprPtr<T>& l, const ExprPtr<T>& r) 
{
    auto predicate = [=]() mutable 
    {
        l.update();
        r.update();
        return value(l) == value(r);
    };
    
    return BooleanExpr(predicate);
}
~~~

autodiff/reverse/var/var.hpp
```diff
@@ -911,19 +1058,168 @@ struct Hypot3Expr : TernaryExpr<T>
...
+    auto operator! () const { return ExpressionComparison<T, Negate<Comparator>>(l, r, Negate<Comparator> {}); }
+};
+
+// CTAD for ExpressionComparison
+template<typename T, typename Comparator>
+ExpressionComparison(const ExprPtr<T>&, const ExprPtr<T>&, Comparator&&) -> ExpressionComparison<T, Comparator>;
+
+/// Capture boolean operations between boolean expressions
+template<typename Operation>
+struct BooleanOperation : public BooleanExpr
```
allanleal, line 1133: Can't we define operators `||`, `&&`, etc. in a way that they produce a new `BooleanExpr`, without creating this additional type, `BooleanOperation`? 

Just trying to see if there is an alternative approach with less indirection. If not possible, then OK.

allanleal: It's possible that I'm overlooking some corner cases here, in which these additional types would provide a more general solution (e.g., mixing `BooleanExpr` with `bool`).

autodiff/reverse/var/var.hpp
jargonzombies: The `wrap` function is a shorthand for generating a shared pointer to base from a derived object instance. I noticed that this pattern came up a lot in the `Expr<T>` free functions and wanted to factor it out. It could be better in principle, a CRTP base could supply it instead, if you prefer.

autodiff/reverse/var/var.hpp
jargonzombies: Are the semantics of copying a lambda instance as you suggest the same as `shared_ptr`? I'm just wondering about the `propagatex` function of `ConditionalExpr<T>` when (currently) the predicate expression tree gets shared with a new expression containing the derivative of the left and right sides of the expression. If, instead, the lambda instance gets copied into the new gradient expression tree, is that okay?

Also, your suggestion would mix update and evaluate functionality, which could be a performance hit in some use cases. It's separated so far for floating expressions, so why mix for booleans?

autodiff/reverse/var/var.hpp
jargonzombies: In principle I could see all of those use cases being encompassed by lambdas. I can try it out if you like, but it would mean a different paradigm for this part of the implementation when compared to the existing bits

autodiff/reverse/var/var.hpp
allanleal: Ah, OK, usually such kind of method is called `clone` instead of `wrap` (at least from my experience). No need for CRTP here (I also avoided it for the forward types `dual` and `real` -- the implementation there relies heavily on SFINAE and `if constexpr` though, but I find it way more intuitive. Looking forward to concepts in C++20!

autodiff/reverse/var/var.hpp
allanleal: The lambda functions would be copied along with their captured expression pointers (`ExprPtr`). 

> Also, your suggestion would mix update and evaluate functionality, which could be a performance hit in some use cases. It's separated so far for floating expressions, so why mix for booleans?

Do you mean the `l.update()` and `r.update()` above and then the `value(l)` and `value(r)`? These `value` calls are cheap - just extract the value of the evaluated expressions `l` and `r`. Is there a way to avoid this? My feeling is that these expressions `l` and `r` always need to be evaluated in case there are nodes (independent variables) in these expressions whose values changed.

autodiff/reverse/var/var.hpp
allanleal: I'm not totally sure that lambda functions would eliminate the need for these additional types, but they are powerful constructs not to be underestimated. If this becomes too much work, however, let's go with your current approach.

autodiff/reverse/var/var.hpp
jargonzombies: The reason I'm worried is that an `update` call at the root of an expression tree is a non-tail-recursive DFS-like traversal of the entire expression tree with a stack frame per layer without assurances that nodes won't be visited twice if there's structural sharing. I haven't benchmarked anything, but that sounds like a potential problem to me.

Changing the implementation like this by mixing `update` and evaluation doesn't really change much since `update` still has to be called at the root of the expression tree by hand due to the value caching. Changing `value` calls to auto-`update` in all expressions would imply that there's no point to caching anymore.

jargonzombies:
You were right, lambdas can do it all. It's probably not worth spending too much time speculating on the performance implications since the boolean part of the conditional expressions is likely to be comparatively simple. If it ever becomes a problem, someone will let you know :wink:

allanleal:
Wow, this looks great! Yes, lambda functions are powerful, and I use them all the time. I'm glad they really solved this problem in a more straightforward way. Jan, I know you've been done already quite some work on this PR, but may I just request from you one example under `autodiff/examples/reverse` and then add a link to this example file in `autodiff/docs/website/tutorials.md`? The `conditional` feature is an excellent addition to the project and it would be great if users could learn about it via an example.

allanleal:
Great. Are you planning for more work on this PR or can I merge this?

jargonzombies:
I think that's all!

Issue #160, opened 2021-06-04
A possible memory issue?


sergeigribanov:
I would like to use autodiff in my project. The project is related to the problem of nonlinear constrained optimization. In the case of my problem, the optimization is done not once, but hundreds of thousands or millions of times in a row. The maximum dimension of the parameter space can be on the order of several hundred. I want to use autodiff to calculate the gradient and hessian of an objective function in an optimization algorithm. Before using autodiff in my project, I decided to conduct a series of tests. In one of my tests, I came across a memory usage issue. The code of the test program is the following:
```c++
// C++ includes
#include <iostream>

// Eigen includes
#include <Eigen/Core>
using namespace Eigen;

// autodiff include
#include <autodiff/reverse.hpp>
#include <autodiff/reverse/eigen.hpp>
using namespace autodiff;

var f(const VectorXvar& x)
{
  return x.dot(x);
}

int main() {
  const int n = 100;
  const int numberOfEntries = 1000000;
  for (int i = 0; i < numberOfEntries; ++i) {
    // VectorXvar x = VectorXvar::Random(n);
    VectorXvar x = VectorXvar::Ones(n);
    var u = f(x);
    VectorXd g;
    MatrixXd H = hessian(u, x, g);
  }
  std::cout << "Finished!" << std::endl;
  return 0;
}
```
CMake file that I use to build this test program is the following:
```
cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
project(autodiff-tests)

find_package(Eigen3 REQUIRED NO_MODULE)
find_package(autodiff)

add_executable(reverse-hessian ${CMAKE_CURRENT_SOURCE_DIR}/src/reverse-hessian.cpp)
target_link_libraries(reverse-hessian Eigen3::Eigen autodiff::autodiff)
```

The issue happens when the variable `numberOfEntries` is equal to 1M. In this case, memory usage gradually rises to 100%. Eventually, the process ends with the message: Killed. 

If the variable `numberOfEntries` is equal to 100K, then the program runs without issues.

I have tested this program on my laptop (16 GB of RAM, CPU:  intel core i7, OS: Arch Linux).

allanleal:
Thanks for reporting this. Are you able to create a reproducible example that does not involve Eigen? This could simplify the memory leaking checking.

jargonzombies:
[...]

PR #177, opened 2021-08-19
Remove grad and gradx members from variable exprs

jargonzombies:
Implementation of an idea from #160 removing `grad` and `gradx` members from `VariableExpr` in favor of value and expression pointers to which updates can be written during expression tree traversal.

Fixes the associated regressions in derivative calculation, too. The tests no longer leak memory.

Some details:
- In principle, there's no need to add `bind_value` and `bind_expr` to the `Expr<T>` interface. For instance `expr.bind_value(nullptr)` can be replaced with a `static_cast<VariableExpr<T>*>(expr.get())->gradPtr = nullptr;` because this is only ever called on variable expressions. Any preference you might have there is easy to adopt.
- The `nullptr` checking in `VariableExpr<T>::propagate` and `::propagatex` could just be replaced with an `assert` for debug build checking if you have runtime performance concerns.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -49,7 +49,7 @@ jobs:
         shell: powershell
         run: |
           conda activate autodiff
-          cmake -S . -B .build
+          cmake -S . -B .build -DAUTODIFF_TEST_SANITIZE=ON
       - name: Build
         shell: powershell
         run: |
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -49,12 +49,12 @@ jobs:
         shell: powershell
         run: |
           conda activate autodiff
-          cmake -S . -B .build
+          cmake -S . -B .build --config ${{ env.configuration }} -DAUTODIFF_TEST_SANITIZE=ON
       - name: Build
         shell: powershell
         run: |
           conda activate autodiff
-          cmake --build .build --config ${{ env.configuration }} -DAUTODIFF_TEST_SANITIZE=ON
+          cmake --build .build --config ${{ env.configuration }}
       - name: Install
         shell: powershell
         run: |
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -64,4 +64,4 @@ jobs:
         shell: powershell
         run: |
           conda activate autodiff
-          cmake --build .build --config ${{ env.configuration }} --target tests
+          cmake --build .build --config ${{ env.configuration }} --target test
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -49,7 +49,7 @@ jobs:
         shell: powershell
         run: |
           conda activate autodiff
-          cmake -S . -B .build -DAUTODIFF_TEST_SANITIZE=ON
+          cmake -S . -B .build
       - name: Build
         shell: powershell
         run: |
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -64,4 +64,4 @@ jobs:
         shell: powershell
         run: |
           conda activate autodiff
-          cmake --build .build --config ${{ env.configuration }} --target test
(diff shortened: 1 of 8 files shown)


autodiff/reverse/var/var.hpp
```diff
@@ -254,6 +255,9 @@ struct Expr
     /// Destructor (to avoid warning)
     virtual ~Expr() {}
 
+    virtual void bind_value(T* /* grad */) {}
```
allanleal, line 258: Please add a short (Doxygen ///) comment here for these `bind_*` functions. I like your approach.

allanleal (commented):
Hi Jan, thanks a lot for the fix. In the end you managed to find a solution that required very little changes. Very good. I have just kindly requested you to add a short comment to two methods below, just to make it clear what are their roles. I've added a memory leak check in the CI.

jargonzombies:
I'm glad you approve! 

By the way, you could cut `valgrind` out of the CI and use sanitizers instead, with something like the following in `tests/CMakeLists.txt`:

```cmake
if(AUTODIFF_SANITIZE_TESTS)
  target_compile_options(autodiff-cpptests PRIVATE "-fsanitize=address,undefined")
  target_link_options(autodiff-cpptests PRIVATE "-fsanitize=address,undefined")
endif()
```

allanleal:
Interesting. I'm OK with either approach. If the CI fails again, could you
please introduce this then (the AUTODIFF_SANITIZE_TESTS logic)?

>

allanleal:
Setting `AUTODIFF_SANITIZE_TESTS` to `ON` in the macOS build is failing. I'll probably enable this only for Linux, which should be enough.

jargonzombies:
Yeah, I think the `cmake` on osx doesn't much like your use of a target named `tests`. If you don't mind, I'm gonna try to fix that.

MSVC did [add Asan](https://devblogs.microsoft.com/cppblog/addresssanitizer-asan-for-windows-with-msvc/) a while ago, so I thought it might be worth trying via the compiler flag testing. Not sure what's been failing the CI pipeline though.

.github/workflows/windows.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         shell: powershell
         run: |
           conda activate autodiff
-          cmake -S . -B .build
+          cmake -S . -B .build -DAUTODIFF_TEST_SANITIZE=ON
```
allanleal, line 52: I don't think this works in Windows. I'll remove this (the only thing missing to get CI fully passing)

adam-ce:
hi,
i'm using autodiff reverse mode to unit test my own manual gradient derivatives for complex functions.

I do functions similar to the following
`template <typename scalar_t, int DIMS>
EXECUTION_DEVICES glm::mat<DIMS, DIMS, scalar_t> inverse(const glm::mat<DIMS, DIMS, scalar_t>& cov, const glm::mat<DIMS, DIMS, scalar_t>& incoming_grad);`
which i want to unit test.
The thing is, that i have several in- and outputs, and i'm manually `var.expr->propagate(grad)` on each of the components and then reading the result from the input vars one by one in a loop. that's now not possible any more. i think that the easiest workaround is to create my own autograd variable and call bind myself.

I understand the reasons for the changes, but if i understand it correctly, it could have been limited to the gradx case. imo it's also a nicer design, if you can get the gradient from the variable directly. anyway, what i actually want to say, is, that it would be nice if you also consider such use cases, i.e., don't try to completely hide the implementation, because that makes it harder to adapt for certain use cases (another example is, that some important functionality is now in detail, which also broke my code).

allanleal:
Hi Adam, many thanks for reporting this. You're right that the memory leak issue was caused only when `gradx` and `propagatex` were used (i.e., for higher-order derivatives).

I would accept changes that revert the `grad` and `propagate` methods. I won't be available for this soon. Would you kindly contribute with a PR that introduces back `grad` and `propagate`?

Issue #178, opened 2021-08-23
pow(x,2) expressions produce NaNs in reverse mode Hessian


c-renton:
Expressions containing `pow(x,2)` can produce NaNs when computing a Hessian in reverse mode.
In the example of computing the gradient and Hessian for the Rosenbrock function below, the computed gradient is correct, but the computed Hessian is 2 by 2 matrix of NaNs.
Changing the point of evaluation away from `(0,0)` or `(1,1)` produces the correct Hessian.
Changing the expression to use `x*x` instead of `pow(x,2)` produces the correct Hessian at all points.
The same issue does not exist when computing the Hessian in forward mode.

```#include <iostream>
#include <cmath>
#include <Eigen/Core>
#include <autodiff/reverse/var.hpp>
#include <autodiff/reverse/var/eigen.hpp>

template<typename Scalar>
static Scalar rosenbrock(const Eigen::Matrix<Scalar,2,1> &x)
{   
    using std::pow; // In case Scalar is float or double
    // Scalar xSq = x(0)*x(0);
    // Scalar yMinusXSq = x(1) - xSq;
    // Scalar OneMinusX = 1 - x(0);
    // return (OneMinusX*OneMinusX + 100*yMinusXSq*yMinusXSq);

    // The following expression fails to produce a correct Hessian
    // at (0,0) or (1,1) using reverse mode (result contains NaNs)
    return (pow(1 - x(0),2) + 100*pow(x(1) - x(0)*x(0),2));
}

int main(int argc, char *argv[])
{
    Eigen::Vector2d x;
    x(0) = 0;
    x(1) = 0;

    Eigen::Matrix<autodiff::var,2,1> xvar = x.cast<autodiff::var>();
    autodiff::var fvar = rosenbrock(xvar);
    Eigen::Vector2d gvar;
    Eigen::Matrix2d Hvar;
    Hvar = hessian(fvar, xvar, gvar);
    std::cout << "f(" << xvar(0) << "," << xvar(1) << ") = " << fvar << std::endl;
    std::cout << "Gradient at (" << xvar(0) << "," << xvar(1) << "):" << std::endl;
    std::cout << gvar << std::endl;
    std::cout << "Hessian at (" << xvar(0) << "," << xvar(1) << "):" << std::endl;
    std::cout << Hvar << std::endl;

    return 0;
}
```

allanleal:
Hi @c-renton 

[...]

PR #180, opened 2021-08-23
Avoid division in reverse-mode pow

c-renton:
Attempt to address https://github.com/autodiff/autodiff/issues/178

### DIFF
--- a/autodiff/reverse/var/var.hpp
+++ b/autodiff/reverse/var/var.hpp
@@ -749,9 +749,9 @@ struct PowExpr : BinaryExpr<T>
     {
         const auto lval = l->val;
         const auto rval = r->val;
-        const auto aux = wprime * val;
-        l->propagate(aux * rval / lval);
-        r->propagate(aux * log(lval));
+        const auto aux = wprime * pow(lval, rval - 1);
+        l->propagate(aux * rval);
+        r->propagate(aux * lval * log(lval));
     }
 
     virtual void propagatex(const ExprPtr<T>& wprime)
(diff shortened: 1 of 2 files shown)


autodiff/reverse/var/var.hpp
```diff
@@ -749,9 +749,9 @@ struct PowExpr : BinaryExpr<T>
     {
         const auto lval = l->val;
         const auto rval = r->val;
-        const auto aux = wprime * val;
-        l->propagate(aux * rval / lval);
-        r->propagate(aux * log(lval));
+        const auto aux = wprime * pow(lval, rval - 1);
+        l->propagate(aux * rval);
+        r->propagate(aux * lval * log(lval));
```
allanleal, line 754: Did you check this work? I'm afraid this will produce NaN as well, because of `log(lval)`.

allanleal: Maybe we need something like this:

~~~
const auto auxr = lval == 0.0 ? 0.0 : lval * log(lval);
 r->propagate(aux * auxr);
~~~

c-renton: It passes the failing test I added and works in the rosenbrock example https://github.com/autodiff/autodiff/issues/178#issue-976690033. It's no worse than the existing expressions in the `propagatex` functions on lines 761 and 782, which also take the log of `lval`.

c-renton: It could be that since I was only looking at `pow(x,y)` for constant `y`, the expression for the derivative w.r.t. `y` is not being propagated.

allanleal:
Thanks a lot for this, @c-renton!

PR #184, opened 2021-09-01
Adds support and examples for using Eigen::Map

jjaraalm:
Allows, `gradient()`, `jacobian()`, and `hessian()` to write to pre-allocated storage. Examples and tests use `Eigen::Map` to write to a POD double array, but any object which implements `resize(n, m)`, `rows()`, `cols()` and `operator(i, j)` can be used.


allanleal (commented):
All looks very good. Thanks for this. Just waiting for the CI to check the tests.

PR #187, opened 2021-09-22
Allow sfinae specialization on VectorTraits

artivis:
[Proposal]

This small change allows for sfinea-like specialization of the `VectorTraits` struct, e.g.

```cpp
struct Foo {};

struct Bar : Foo {};
struct Blarg : Foo {};

namespace autodiff::detail {

/// Specialize VectorTraits for Foo & derived.
template <typename T> struct VectorTraits<T, EnableIf<std::is_base_of_v<Foo, T>>> { ... };

}
```

Thus offering more flexibility in the specialization.

Other traits may benefit from this feature as well but so far I did not needed to change any other to add support to `autodiff` in my project.


allanleal:
Sounds good to me. Thanks @artivis !

Issue #191, opened 2021-10-29
Eigenvalues derivatives and status of the project


giacomo-b:
First of all, keep up the great work! I think this library is awesome and I'd like to contribute one day.

I have a matrix `M` whose entries depend on one or more input parameters, and I am computing the eigenvalues of `M` using [spectra](https://github.com/yixuan/spectra). What I would like to achieve using autodiff is the computation of the derivatives of these eigenvalues w.r.t. the input parameters (basically, their sensitivities). Since spectra supports custom types I thought it might be possible to have it work seamlessly with autodiff.

The problem I am facing is that I don't know:
1) if this is possible using autodiff
2) assuming it can be done, which types I should use to achieve this.

I went through the [tutorials](https://autodiff.github.io/tutorials/) but didn't find anything similar.

The steps are:
1. Choose a parameter `p`
2. Compute a matrix `M` as a function of `p`
3. Compute the eigenvalues `eigs` of `M` (done using spectra)
4. Compute the derivatives of `eigs` w.r.t. `p` (which of course go through the computation of `M` and the eigensolver)

Simple example for one input parameter and a 2x2 matrix:

```cpp
// Given a parameter p, compute the M(p)
// Should probably return an autodiff type (MatrixXreal?) and take a const ref as argument
Eigen::MatrixXd M(double p)
{
    Eigen::MatrixXd output;
    output << 2 * p,  p * p,
              4 * p,    p;
    return output;
}

// Use spectra to compute the eigenvalues
Eigen::VectorXd eigs = ... // The eigenvalues should probably be saved in a VectorXreal

// Compute sensitivities using autodiff (d(eigs) / dp)
// ?
```

Do you think this is possible? If so, could you please show me how or point me to some related examples?

Also, quick question: what is the current status of the project? I noticed that the release version is not 1.0 yet, so I was wondering if there are any specific features in the pipeline that are currently not supported but you think are worth mentioning.

Thanks!

allanleal:
Great, keep me posted!

mattarroz:
Hi @allanleal,

thumbs up for autodiff, great piece of software!

> This should be possible. I've started creating an example here, but noticed that some things will need to be implemented (e.g., Eigen uses `std::isfinite` on the number type of the matrix when evaluating eigenvalues, and this is causing compilation errors).

[...]

PR #193, opened 2021-11-17
Added T cast operator to Real

mattarroz:
The cast operator is needed to make `std::isfinite` work on `Real`, as discussed in https://github.com/autodiff/autodiff/issues/191#. In consequence, it is now possible to compute derivatives for Eigenvalue/vector computations.

### DIFF
--- a/autodiff/forward/real/real.hpp
+++ b/autodiff/forward/real/real.hpp
@@ -95,6 +95,11 @@ public:
         return m_data[0];
     }
 
+    operator T() const
+    {
+        return (T)m_data[0];
+    }
+
     constexpr auto operator[](size_t i) -> T&
     {
         return m_data[i];


mattarroz:
I had tested autodiff only as a header only-library and wasn't aware of the compile errors. Unsure if the fixes to min and max are correct.

autodiff/forward/real/real.hpp
```diff
@@ -797,13 +802,13 @@ constexpr auto min(const Real<N, T>& x, const Real<N, T>& y)
 template<size_t N, typename T, typename U, EnableIf<isArithmetic<U>>...>
 constexpr auto min(const Real<N, T>& x, const U& y)
 {
-    return (x[0] <= y) ? x : y;
+    return (x[0] <= y) ? x[0] : y;
```
PaulXiCao, line 805: Your change will result in a _different return type_. Previously `min` returned `Real<N, T>` and now some mix of `T` and `U`, e.g. if both are of type `double` then it will just return a `double` value! Thus, you truncate all the derivatives.

autodiff/forward/real/real.hpp
```diff
@@ -815,13 +820,13 @@ constexpr auto max(const Real<N, T>& x, const Real<N, T>& y)
 template<size_t N, typename T, typename U, EnableIf<isArithmetic<U>>...>
 constexpr auto max(const Real<N, T>& x, const U& y)
 {
-    return (x[0] >= y) ? x : y;
+    return (x[0] >= y) ? x[0] : y;
```
PaulXiCao, line 823: Same comment as above. We probably don't want this change.

tests/forward/real/real.test.cpp
```diff
@@ -381,16 +381,16 @@ TEST_CASE("testing autodiff::real", "[forward][real]")
...
     z = max(x, y);    CHECK( z == y );
     z = max(y, x);    CHECK( z == y );
-    z = max(x, 0.1);  CHECK( z == x );
-    z = max(0.2, x);  CHECK( z == x );
-    z = max(0.5, x);  CHECK( z == x );
-    z = max(x, 0.5);  CHECK( z == x );
+    z = max(x, 0.1);  CHECK( z == x[0] );
+    z = max(0.2, x);  CHECK( z == x[0] );
+    z = max(0.5, x);  CHECK( z == x[0] );
+    z = max(x, 0.5);  CHECK( z == x[0] );
```
PaulXiCao, line 393: We probably don't want this change. See previous comment.

autodiff/forward/real/real.hpp
```diff
@@ -95,6 +95,11 @@ class Real
         return m_data[0];
     }
 
+    operator T() const
+    {
+        return (T)m_data[0];
```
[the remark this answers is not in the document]
mattarroz: Fixed.

autodiff/forward/real/real.hpp
```diff
@@ -797,13 +802,13 @@ constexpr auto min(const Real<N, T>& x, const Real<N, T>& y)
...
 {
-    return (x[0] <= y) ? x : y;
+    return (x[0] <= y) ? x[0] : y;
 }
 
 template<size_t N, typename T, typename U, EnableIf<isArithmetic<U>>...>
 constexpr auto min(const U& x, const Real<N, T>& y)
 {
-    return (x < y[0]) ? x : y;
+    return (x <  y[0]) ? x : y[0];
```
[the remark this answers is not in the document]
mattarroz: Fixed.

autodiff/forward/real/real.hpp
```diff
@@ -815,13 +820,13 @@ constexpr auto max(const Real<N, T>& x, const Real<N, T>& y)
...
 {
-    return (x[0] >= y) ? x : y;
+    return (x[0] >= y) ? x[0] : y;
 }
 
 template<size_t N, typename T, typename U, EnableIf<isArithmetic<U>>...>
 constexpr auto max(const U& x, const Real<N, T>& y)
 {
-    return (x > y[0]) ? x : y;
+    return (x > y[0]) ? x : y[0];
```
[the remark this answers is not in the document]
mattarroz: Fixed.

autodiff/forward/real/real.hpp
mattarroz: Feel free to have a look at my solution which leaves the interface unchanged. Maybe there is a better way? 

With the added T() operator, the method would not compile as it was previously. By the way, the same error happens if AUTODIFF_ENABLE_IMPLICIT_CONVERSION_REAL is defined.

PaulXiCao: Your change to `return (x[0] <= y) ? x : Real<N, T>(y);` seems useful! Thanks!

autodiff/forward/real/real.hpp
PaulXiCao: Fixed! Thanks :+1:

tests/forward/real/real.test.cpp
PaulXiCao: Fixed! Thanks :+1:

autodiff/forward/real/real.hpp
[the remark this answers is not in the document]
PaulXiCao: I just thought about it again and there is actually no need to do a type conversion at all. `m_data[0]` is already of type `T`! Would you mind removing the conversion again?

tests/forward/real/real.test.cpp
```diff
@@ -409,6 +409,8 @@ TEST_CASE("testing autodiff::real", "[forward][real]")
     CHECK_APPROX( z[3], std::abs(y[0])/(y[0]) * y[3] );
     CHECK_APPROX( z[4], std::abs(y[0])/(y[0]) * y[4] );
 
+    CHECK(std::isfinite(y[0]) == std::isfinite(y));
```
PaulXiCao, line 412: The test seems useful. Could you also add explicit tests for the introduced `T()` operator, i.e. one for `double` and one for `float`?

autodiff/forward/real/real.hpp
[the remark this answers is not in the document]
PaulXiCao: Done. Thanks

tests/forward/real/real.test.cpp
PaulXiCao: Done. Thanks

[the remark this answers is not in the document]
mattarroz: (see resolved conversation below)

PR #207, opened 2022-03-10
Fix some compiler warnings

rabraker:
This PR corrects some compiler warnings:

- Removes an unused variable from `ForEachWrtVar` in `forward/utils/gradient.hpp`
- Fixes some warnings about `-Wsign-compare`, stemming from a mismatch in types between the loop index and a some size variable in several places.

After this change, I can compile cleanly with `-Wall -Werror`

### DIFF
--- a/autodiff/forward/real/real.hpp
+++ b/autodiff/forward/real/real.hpp
@@ -854,7 +854,7 @@ auto repr(const Real<N, T>& x)
 {
     std::stringstream ss;
     ss << "autodiff.real(";
-    for(auto i = 0; i <= N; ++i)
+    for(decltype(N) i = 0; i <= N; ++i)
         ss << (i == 0 ? "" : ", ") << x[i];
     ss << ")";
     return ss.str();
--- a/autodiff/forward/utils/derivative.hpp
+++ b/autodiff/forward/utils/derivative.hpp
@@ -123,7 +123,8 @@ auto seed(const At<Args...>& at, const Along<Vecs...>& along)
         if constexpr (isVector<decltype(arg)>) {
             static_assert(isVector<decltype(dir)>);
             assert(arg.size() == dir.size());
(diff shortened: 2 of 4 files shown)


autodiff/forward/real/real.hpp
```diff
@@ -854,7 +854,7 @@ auto repr(const Real<N, T>& x)
 {
     std::stringstream ss;
     ss << "autodiff.real(";
-    for(auto i = 0; i <= N; ++i)
+    for(decltype(N) i = 0; i <= N; ++i)
```
PaulXiCao, line 857: There are other "styles" possible here, e.g.
* `for(auto i = 0u; ...`
* `for(std::size_t i = 0; ...`

Which is the preferred style for autodiff?

autodiff/forward/utils/derivative.hpp
```diff
@@ -123,7 +123,8 @@ auto seed(const At<Args...>& at, const Along<Vecs...>& along)
         if constexpr (isVector<decltype(arg)>) {
             static_assert(isVector<decltype(dir)>);
             assert(arg.size() == dir.size());
-            for(auto i = 0; i < dir.size(); ++i)
+            size_t len = dir.size();
+            for(decltype(len) i = 0; i < len; ++i)
```
PaulXiCao, line 127: One could remove `len` if we would use `auto i = 0u` or something similar.

PaulXiCao (commented):
Fixing compiler warnings is always a good deed! :)

PaulXiCao:
Starting from cmake 3.24 there is also a target property for it: [COMPILE_WARNING_AS_ERROR](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_WARNING_AS_ERROR.html#prop_tgt:COMPILE_WARNING_AS_ERROR).
Is there anything holding us back to cmake 3.16 which is currently the minimal required version?

PR #214, opened 2022-03-23
Fix "unexpected end-of-file" error in C++20

Arsennnic:
Fixed the error raised by MSVC (v143) C++20 compiler:
```
autodiff/forward/utils/derivative.hpp(107,1): fatal error C1004: unexpected end-of-file found
```


allanleal:
Thanks!

PR #224, opened 2022-04-19
Compute Jacobian with respect to a matrix

robertodr:


artivis:
Hi @robertodr, 
What was the point of this PR? There is no description whatsoever. FYI it introduced a breaking change #233.

Issue #185, opened 2021-09-20
Missing atan2 for type real


artivis:
The function `atan2` is not overloaded for the `autodiff::real` type.

allanleal:
Hi @artivis - would you kindly work on this? Here are my notes on the higher-order derivative expressions for `real`:

[higher-order-derivatives-formulas.pdf](https://github.com/autodiff/autodiff/files/7200296/higher-order-derivatives-formulas.pdf)

Unfortunately, these are only notes. I haven't published them and they are not explained in-depth. You should though be able to infer from there how to derive the expression for `tan2` and maybe the missing others.

bostoncleek:
I am working on a PR to add `atan2`. The picture below contains the derivations for when `atan2` is used with a scalar and a dual number. I have been using PR https://github.com/autodiff/autodiff/pull/116 as a reference.

I am struggling to derive the equations for `atan2(y,x)` when both `x` and `y` are dual numbers. @allanleal 

![atan2](https://user-images.githubusercontent.com/36201047/165611004-aed195a7-9151-45d3-a2cb-3d99aff78986.png)

bostoncleek:
closed y https://github.com/autodiff/autodiff/pull/226

PR #226, opened 2022-04-29
Add atan2(y,x) to autodiff::Real

bostoncleek:
This PR adds `atan2(y,x)` to `autodiff::Real`.

Related issue https://github.com/autodiff/autodiff/issues/185

### DIFF
--- a/autodiff/forward/real/real.hpp
+++ b/autodiff/forward/real/real.hpp
@@ -207,6 +207,7 @@ using std::acos;
 using std::acosh;
 using std::asin;
 using std::asinh;
+using std::atan2;
 using std::atan;
 using std::atanh;
 using std::cbrt;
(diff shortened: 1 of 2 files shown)


autodiff/forward/real/real.hpp
```diff
@@ -656,6 +657,60 @@ constexpr auto atan(const Real<N, T>& x)
...
+
+template<size_t N, typename T>
+constexpr auto atan2(const Real<N, T>& y, const Real<N, T>& x)
+{
+    Real<N, T> res;
+    res[0] = atan2(y[0], x[0]);
+    if constexpr(N > 0) {
+        const T denom = x[0] * x[0] + y[0] * y[0];
+        For<1, N + 1>([&](auto i) constexpr {
+            res[i] = (x[0] * y[i] - x[i] * y[0]) / denom;
```
bostoncleek, line 708: @allanleal I am not entirely sure why this is true and was not able to derive it for myself. I mainly copied the pattern from  the implementation for `dual` seen [here](https://github.com/autodiff/autodiff/blob/main/autodiff/forward/dual/dual.hpp#L1403-L1407). If you could help me understand why this is valid I would greatly appreciate it.

allanleal: Thanks, @bostoncleek . I haven't yet published the algorithm behind `real` (just don't have time now). It's an algorithm different than the classic dual approach, specialized for single-order and higher-order directional derivatives.

bostoncleek: I'm very interested in the paper and will be awaiting its release. Are there similar papers you can refer me to or are you saying its a completely different algorithm?

Issue #223, opened 2022-04-10
Cannot compile with Cmake 3.16.3


jefersonjlima:
Hello,

I get some errors when I compile `autodiff` with `Cmake - version 3.16.3`. The `CMakeLists.txt` has the `cmake_minimum_required(VERSION 3.0)`, but, I think it isn't right. 

I compile with `Cmake 3.23.0` and it works fine.

allanleal:
Hi @jefersonjlima . I was on vacation. Thanks for reporting this. Could you please let me know the exact error? With this info, I'll be able to identify the minimum cmake version (which is certainly not 3.0 indeed).

jefersonjlima:
Hi @allanleal , I'm sorry about the delay.
I retest this issue in another machine and got the same error:

```
user:/tmp/autodiff/.build$ cmake --build . --target install
Scanning dependencies of target autodiff-cpptests
[  1%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/main.cpp.o
[  3%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/common/meta.test.cpp.o
[  4%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/forward/dual/dual.test.cpp.o
[  6%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/forward/dual/eigen.test.cpp.o
[  7%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/forward/real/eigen.test.cpp.o
[  9%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/forward/real/real.test.cpp.o
[ 10%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/forward/utils/derivative.test.cpp.o
[ 12%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/forward/utils/gradient.test.cpp.o
[ 13%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/forward/utils/taylorseries.test.cpp.o
[ 15%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/reverse/var/eigen.test.cpp.o
[ 16%] Building CXX object tests/CMakeFiles/autodiff-cpptests.dir/reverse/var/var.test.cpp.o
[ 18%] Linking CXX executable autodiff-cpptests
[ 18%] Built target autodiff-cpptests
Scanning dependencies of target autodiff4py
[...]

PR #228, opened 2022-05-11
Introduces potential pybind11 ABI fix

allanleal:
The python bindings of `autodiff` produced with `pybind11` only work well with python bindings of other codes if the same compiler (and same version!) are the same. Otherwise, runtime errors happen (e.g., not possible to do implicit conversion from `float` to `real`). 

This PR introduces the changes recommended here (https://github.com/pybind/pybind11/pull/2602) to let pybind11 to be less strict on how it determines ABI compatibility.

This PR also fixes the cmake version issue reported in #223 . 

This PR also considerably improves compilation time for the python bindings.


ianhbell:
I ran into what I believe to be a similar or same issue with pybind11 and reported it a LONG time ago: https://github.com/pybind/pybind11/issues/461

ianhbell:
Hmm maybe it wasn't that issue

PR #236, opened 2022-07-15
CMake: Avoid check of architecture.

marscho1:
autodiff is a header only library. Therefore no check of architecture is required. If one does not disable this, CMake will complain if autodiff was built in 64 bit but later on is used by a 32 bit application.


allanleal:
Thanks, @marscho1 . I was on vacation entire July.

PR #256, opened 2022-09-26
cpp-20 lambda fix

seba2390:
Ensuring cpp-20 compiler standard compatibility by fixing reference in lambda function.

### DIFF
--- a/autodiff/reverse/var/var.hpp
+++ b/autodiff/reverse/var/var.hpp
@@ -1096,7 +1096,7 @@ struct BooleanExpr
 
     void update() { val = expr(); }
 
-    auto operator! () const { return BooleanExpr([=]() { return !(expr()); }); }
+    auto operator! () const { return BooleanExpr([=, this]() { return !(expr()); }); }
 };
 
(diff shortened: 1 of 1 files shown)


autodiff/reverse/var/var.hpp
```diff
@@ -1096,7 +1096,7 @@ struct BooleanExpr
 
     void update() { val = expr(); }
 
-    auto operator! () const { return BooleanExpr([=]() { return !(expr()); }); }
+    auto operator! () const { return BooleanExpr([=, this]() { return !(expr()); }); }
```
PaulXiCao, line 1099: It might be enough to capture `this`, i.e. `[this](){ return ...}`.

PaulXiCao:
Just for reference: see here for the documentation of lambda-captures: [cppreference.com](https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture)

PaulXiCao (changes requested):

PaulXiCao:
If C++20 is a goal then we might should search for similar things. A simple regex might be enough at first or parse the code base with libclang?

PR #260, opened 2022-10-17
Enable CMake integration with FetchContent

### LINKED ISSUE #259: Enable CMake integration with FetchContent
It would be very convenient to enable CMake integration with FetchContent like so
```
include(FetchContent)
FetchContent_Declare(
        autodiff
        GIT_REPOSITORY https://github.com/autodiff/autodiff
        GIT_TAG        v0.6.12)
FetchContent_MakeAvailable(autodiff)
```

jeanchristopheruel:
Integrate autodiff in CMake using FetchContent
```
include(FetchContent)
FetchContent_Declare(
        autodiff
        GIT_REPOSITORY https://github.com/autodiff/autodiff
        GIT_TAG        v0.6.12)
FetchContent_MakeAvailable(autodiff)
```
And for more granular control:
```
include(FetchContent)
FetchContent_Declare(
        autodiff
        GIT_REPOSITORY https://github.com/autodiff/autodiff
        GIT_TAG        v0.6.12)
FetchContent_GetProperties(autodiff)
if(NOT autodiff_POPULATED)
    FetchContent_Populate(autodiff)

    # Bring the populated content into the build
    add_subdirectory(${autodiff_SOURCE_DIR} ${autodiff_BINARY_DIR})
    set(AUTODIFF_BUILD_TESTS OFF)
    set(AUTODIFF_BUILD_PYTHON OFF)
    set(AUTODIFF_BUILD_EXAMPLES OFF)
    set(AUTODIFF_BUILD_PYTHON OFF)
endif()
```

Fixes #259


allanleal:
Hi @jeanchristopheruel , this is how I think we should do it. All CMake changes related to `FetchContent` should exist in a CMake module within the [`cmake`](https://github.com/autodiff/autodiff/tree/main/cmake) directory. Say, `EnableFetchContent.cmake`. Then, in the main/root `CMakeLists.txt` file, we have:

~~~
if(AUTODIFF_ENABLE_FETCH_CONTENT)
    include(EnableFetchContent)
endif()
~~~

I see that you have changed options from `ON` to `OFF` (build tests, python, docs, etc) because if you are integrating `autodiff` with your lib, you probably don't need to build that. But in development mode, I'd hate to have to set these options explicitly all the time! :) Thus, please set them OFF in the `EnableFetchContent.cmake`. 

Also, I will not rely on this fetching mechanism of CMake for installing dependencies. I rely on conda (conda-forge more exactly)  as a dependency manager, which you should give it a try. 

A common misconception is that conda is for python only; I use conda to install even compilers! Have a look at the [environment.devenv.yml](https://github.com/autodiff/autodiff/blob/main/environment.devenv.yml) file.

jeanchristopheruel:
> Also, I will not rely on this fetching mechanism of CMake for installing dependencies. I rely on conda (conda-forge more exactly) as a dependency manager, which you should give it a try.

Interesting indeed. I'm not that familiar with conda. Personally, I think FetchContent makes the code more portable and avoids adding dependencies. Thank you for your answer, I will suggest an improvement.

nilsvu:
Just leaving a note here that FetchContent works well even without the changes in this PR, like this:

```cmake
  include(FetchContent)
  FetchContent_Declare(autodiff
    GIT_REPOSITORY https://github.com/autodiff/autodiff/
    GIT_TAG v1.1.2
    GIT_SHALLOW TRUE
  )
  set(AUTODIFF_BUILD_TESTS OFF CACHE BOOL "Build autodiff tests")
  set(AUTODIFF_BUILD_EXAMPLES OFF CACHE BOOL "Build autodiff examples")
  set(AUTODIFF_BUILD_PYTHON OFF CACHE BOOL "Build autodiff Python bindings")
  set(AUTODIFF_BUILD_DOCS OFF CACHE BOOL "Build autodiff documentation")
  FetchContent_MakeAvailable(autodiff)
```

This way no dependencies of autodiff are installed, but they are all optional anyway. If you need Eigen then you have to fetch/install it before fetching autodiff.

PR #286, opened 2023-04-20
Fix support for using different types in Dual

yut23:
This PR fixes some compiler errors I encountered when trying to use different (incompatible) types for the value and grad terms in the forward-mode `Dual` class.

My particular use-case is computing the Jacobian of an expensive multi-variable function, with ~10-30 inputs I care about. I use an array type for `grad`, and seed each input variable by setting the corresponding entry in `grad` to 1 and the rest to 0. This lets me get all the derivatives with a single function evaluation, rather than evaluating it once for each of the inputs.


allanleal:
Hi @yut23 , thanks for this. I've enabled CI for your PR and let's check if all tests pass.

yut23:
The bazel failure looks unrelated, as it's been failing on main since July 31.

allanleal:
I'll merge this - I'm not familiar with Bazel (this was a contribution by a user). I'll ignore the bazel build fail. Thanks for working on this PR. Seeding individual input variables that matter is a good strategy.

PR #303, opened 2023-11-04
Improve Bazel support

Vertexwahn:
- Apply buildifier to Bazel build files
- Add Bzlmod `MODULE.bazel` file
- Rename from BUILD to BUILD.bazel
- Switch angle brackets to double qoutes


.arquive/benchmarks/benchmark1.cpp
```diff
@@ -10,7 +10,7 @@ using namespace Eigen;
 
 // autodiff includes
 #define AUTODIFF_ENABLE_EIGEN_SUPPORT
-#include <autodiff/reverse/var.hpp>
+#include "autodiff/reverse/var.hpp"
```
allanleal, line 13: Hi @Vertexwahn , why do we need to change from `<>` to `""`?

allanleal (changes requested):
Hi @Vertexwahn , thanks for your contribution. Can we restrict this Pull Request to changes related to Bazel only? See other question on the need to change `<>` to `""`.

.arquive/benchmarks/benchmark1.cpp
Vertexwahn: @allanleal  I would use angle brackets only for system headers (e.g. `<iostream>`, `<windows.h>`) and double quotes for anything else ([1](https://stackoverflow.com/questions/65700648/c-should-i-use-quotes-or-brackets-to-include-headers-in-a-separate-directory), [2](https://www.sololearn.com/Discuss/916636/is-there-any-difference-between-including-the-header-file-within-angular-braces-or-double-quotes), [3](https://stackoverflow.com/questions/3162030/difference-between-angle-bracket-and-double-quotes-while-including-heade)) - this helps to get rid of `copts = ["-I./"]` which does not work when autodiff is used as a dependency

Vertexwahn: @allanleal Is there a reason why this should or cannot be changed to double quotes? Let me know if this is definitely no option for you.

allanleal: I understand your point. Note that the use of `<>` instead of `""` can be seen in prominent C++ projects (e.g. [Boost](https://github.com/boostorg/algorithm/blob/develop/include/boost/algorithm/algorithm.hpp)). Before making this change, we would need to ensure that the integration of `autodiff` in projects that depend on the CMake `find_package` command will not be broken. Would you be able to check that?

PR #309, opened 2024-01-07
[feat] GPU support 

JieRen98:


JieRen98:
Dual is almost done, please check. Please consider applying for a GPU instance for testing cuda codes from GitHub Resources.

JieRen98:
real is supported

JieRen98:
For me, I only need forward. In addition, reverse is not easy to implement because it uses make_shared.

JieRen98:
How to add cuda-toolkit to the test environment? Usually, I use `conda install cuda-toolkit -c nvidia` to install.

allanleal:
Hi @JieRen98 , thanks a lot for your contribution. Are you planning on more additions to this PR?

JieRen98:
For CUDA functions, I think they are mostly done. Of course, we can discuss.

Allan Leal ***@***.***>于2024年1月8日 周一12:53写道：

JieRen98:
CUDA compile options `--expt-relaxed-constexpr --extended-lambda` are necessary, I think we should also add them to the bazel configuration file.

JieRen98:
Workflows may need to change too.

PR #321, opened 2024-02-10
Adds ::reverse namespace to avoid clash with forward

### LINKED ISSUE #179: Can't include forward and reverse mode headers in same compilation unit
I have existing code using autodiff v0.5.13 that I am migrating to v0.6.3 that uses both forward mode and reverse mode auto-differentiation in the same compilation unit.
In updating to the newer version, it appears no longer possible to include both `var` and `dual` (or `real`) headers in the same compilation unit due to conflicting template declarations, which appears to be a regression from previous versions of the library.
A workaround is to separate code using forward mode from code using reverse mode into different compilation units.

```
#include <autodiff/reverse/var.hpp>
#include <autodiff/forward/dual.hpp>

int main(int argc, char *argv[])
{
    return 0;
}
```

alecjacobson:
Fixes https://github.com/autodiff/autodiff/issues/179

Changes the ::detaill to ::reverse::detail namespace in reverse/var/*.hpp so that forward and reverse can be included in the same compilation unit. 

For example, now this minimal example compiles:

```cpp
#include <autodiff/forward/dual.hpp>
#include <autodiff/forward/real.hpp>
#include <autodiff/reverse/var.hpp>
int main() { return 0; }
```

### DIFF
--- a/autodiff/reverse/var/eigen.hpp
+++ b/autodiff/reverse/var/eigen.hpp
@@ -7,7 +7,7 @@
 //
 // Licensed under the MIT License <http://opensource.org/licenses/MIT>.
 //
-// Copyright (c) 2018-2022 Allan Leal
+// Copyright © 2018-2024 Allan Leal
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


allanleal:
Great, thanks for this.

PR #327, opened 2024-03-20
Make the default constructor for Dual constexpr

yut23:
This is needed for global variables with CUDA memory space specifiers (`__device__`, `__shared__`, `__constant__`, and `__managed__`).


allanleal:
Thanks for the improvements.

PR #336, opened 2024-05-28
Fix analytical derivatives in test

gvilanovac:
In the test example-forward-multi-variable-combine-analytical-derivative.cpp, the analytical x and y derivatives of A were swapped.


allanleal (approved):

allanleal:
Thanks!

PR #350, opened 2024-07-06
Make evaluation order for Dual consistent with other numeric types

### LINKED ISSUE #349: Evaluation order for `autodiff::dual` doesn't match that of normal numeric types
When assigning the result of a multi-way addition or multiplication between `autodiff::dual` values, the operations are grouped differently than if the operands were normal numeric types. For example, `x = a + b + c + d` is evaluated like `x = (((a + b) + c) + d)` for normal types, but like `x = (((d + a) + b) + c)` for `autodiff::dual`. This has been giving us unexpected roundoff diffs as we convert our code from manual analytic derivative calculations to autodiff.

Here's some code I wrote to determine the exact evaluation order:
```cpp
#include <iostream>
#include <string>

#include <autodiff/forward/dual.hpp>

struct VerboseNumber {
    std::string name;

    VerboseNumber& operator+=(const VerboseNumber& rhs) {
        this->name = "(" + this->name + " + " + rhs.name + ")";
        return *this;
    }

    VerboseNumber& operator*=(const VerboseNumber& rhs) {
        this->name = "(" + this->name + " * " + rhs.name + ")";
        return *this;
    }
};

VerboseNumber operator+(VerboseNumber lhs, const VerboseNumber& rhs) {
    lhs += rhs;
    return lhs;
}

VerboseNumber operator*(VerboseNumber lhs, const VerboseNumber& rhs) {
    lhs *= rhs;
    return lhs;
}

template <>
struct autodiff::detail::ArithmeticTraits<VerboseNumber> {
    static constexpr bool isArithmetic = true;
};

template <typename number_t>
VerboseNumber eval_order_test() {
    number_t a = VerboseNumber{"a"};
    number_t b = VerboseNumber{"b"};
    number_t c = VerboseNumber{"c"};
    number_t d = VerboseNumber{"d"};
    number_t x = a + b + c + d;
    return autodiff::val(x);
}

int main() {
    std::cout << "evaluating x = a + b + c + d:\n";
    auto normal_result = eval_order_test<VerboseNumber>();
    auto dual_result = eval_order_test<autodiff::Dual<VerboseNumber, double>>();
    std::cout << "normal: equivalent to x = " << normal_result.name << "\n";
    std::cout << "Dual:   equivalent to x = " << dual_result.name << "\n";
}
```

I've got a PR to fix this on the way.

yut23:
Fixes: https://github.com/autodiff/autodiff/issues/349

### DIFF
--- a/autodiff/forward/dual/dual.hpp
+++ b/autodiff/forward/dual/dual.hpp
@@ -1000,13 +1000,13 @@ AUTODIFF_DEVICE_FUNC constexpr void assign(Dual<T, G>& self, U&& other)
     }
     // ASSIGN AN ADDITION EXPRESSION: self = expr + expr
     else if constexpr (isAddExpr<U>) {
-        assign(self, other.r);
-        assignAdd(self, other.l);
+        assign(self, other.l);
+        assignAdd(self, other.r);
     }
     // ASSIGN A MULTIPLICATION EXPRESSION: self = expr * expr
     else if constexpr (isMulExpr<U>) {
-        assign(self, other.r);
-        assignMul(self, other.l);
+        assign(self, other.l);
+        assignMul(self, other.r);
     }
     // ASSIGN A POWER EXPRESSION: self = pow(expr)
     else if constexpr (isPowExpr<U>) {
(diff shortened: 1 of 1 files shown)


yut23:
Closed in favor of #352.

PR #352, opened 2024-07-12
Optionally disable expression reordering for dual

yut23:
If `AUTODIFF_STRICT_ASSOCIATIVITY` is defined before including autodiff headers, optimizations that break the left-to-right associativity of + and * operators will be disabled. This uses more intermediate temporary assignments, but should produce the exact same values between identical code that uses `autodiff::dual` in place of `double`.

Addresses #349.


allanleal:
Hi @yut23 thanks for this. I'm currently on vacation. Will check this early August.

PR #362, opened 2024-10-24
Export matrices for dual

supersega:


supersega:
@allanleal, please make a review)

PR #365, opened 2024-11-13
Fix optional dependence on Eigen

nilsvu:
The dependence on Eigen is optional, so this change allows the CMake configuration to go through when Eigen is not installed.


allanleal:
Thanks, @nilsvu - let's see what happens with the tests. I think their compilation should be conditioned to the existence of Eigen in the system.

PR #367, opened 2024-11-13
Fix optional dependence on Eigen in gradient.hpp

nilsvu:
Before, this header could only be included if Eigen was found, but it is also useful without Eigen.


allanleal:
Hi @nilsvu - thanks for fixing this. Makes sense to disable those methods that depend on eigen (directly or indirectly).

